PR #1, opened 2024-02-13
Fix repo path

mladjan-gadzic:
Simple fix for repo path.

For some reason cloning or updating Hive did not work - but others worked.

Something I've noticed that could be improved is example of script usage or error message if used wrong.


xBis7:
@mladjan-gadzic Thanks, I'm merging this.

> Something I've noticed that could be improved is example of script usage or error message if used wrong.

Feel free to make the changes and create another PR.

PR #6, opened 2024-03-08
Fix tests and add DROP_PARTITION Hive event test

mladjan-gadzic:


xBis7 (approved):

xBis7:
@mladjan-gadzic Thanks!

PR #7, opened 2024-03-11
Add tests for insert and truncate

mladjan-gadzic:


xBis7:
cc @GeorgeJahad

mladjan-gadzic:
> I've run `test_all.sh` for Spark, with these changes and there have been no errors.

Thanks!

PR #11, opened 2024-03-14
Add tests for Hive URL policy

mladjan-gadzic:


trino-hms-hdfs-ranger/tests/spark/hive_url_policies/1_test_no_hive_url_policies.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
xBis7, line 1: Shouldn't we add this in `test_all.sh`?

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/1_test_no_hive_url_policies.sh
```diff
@@ -0,0 +1,21 @@
...
+./setup/load_ranger_policies.sh "$abs_path" "$HDFS_AND_HIVE_ALL"
+
+echo ""
+echo "- INFO: Ranger policies updated."
+
+echo ""
+echo "- INFO: Create table"
+echo "- INFO: [create] should fail."
+failMsg="Permission denied: user [spark] does not have [WRITE] privilege on [[hdfs://namenode/opt/hive/data/sports, hdfs://namenode/opt/hive/data/sports/]]"
+sql="spark.sql(\\\"create table sports (id int, name string)\\\")"
```
xBis7, line 20: table sports is also created in `3_test_spark_hdfs_hive_all.sh`. Is that the reason why this isn't running in `test_all.sh`? 

Can we replace the table names with variables and add them in `testlib.sh` like the rest? It's so much easier to change a value in 1 place instead of 10, if needed.

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/1_test_no_hive_url_policies.sh
mladjan-gadzic: If those tests were to be added to test_all.sh, one of things would happen

1. tests for Hive URL policies will fail because Hive is not configured to work with Hive URL policies
2. tests without Hive URL policies will fail because Hive is configured to work without Hive URL policies

Adding these tests to test_all.sh is going to be tricky because of Hive URL policies. What I can think of is to create one more test_all.sh. There would be two test_all.sh scripts named test_all_URL_policies and test_all.sh. First would run only tests for Hive URL policies and second would run everything else. What do you think?

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/1_test_no_hive_url_policies.sh
mladjan-gadzic: > table sports is also created in `3_test_spark_hdfs_hive_all.sh`. Is that the reason why this isn't running in `test_all.sh`?
> 
No, the reason is outlined in previous comment.

> Can we replace the table names with variables and add them in `testlib.sh` like the rest? It's so much easier to change a value in 1 place instead of 10, if needed.

Sure, I will do that.

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/1_test_no_hive_url_policies.sh
xBis7: All these tests are supposed to be independent from each other. That's why we load different policies every time.

When we have URL policies configured don't we expect slightly different behavior? 

At the end of the existing tests, we can add these ones. They don't have to happen somewhere in the middle of the existing ones. They will load their policies and try to create tables with new names. That shouldn't affect the existing tables.

What am I missing?

xBis7: > At the end of the existing tests, we can add these ones. 

Or at the beginning like you said. I think that as long as we load the correct policies and give them new names that don't already exist, we should be fine, adding them to the existing `test_all.sh`

mladjan-gadzic: One thing that is missing is that Hive needs to be configured to use URL policies or not. Before tests even start to execute. It's not about setting up policies, it is about configuring Hive before starting it.

mladjan-gadzic: There is property `ranger.plugin.hive.urlauth.filesystem.schemes` that needs to have `file:` value and should be set in `ranger-hive-security.xml` in Hive repo.

xBis7: ooh, I forgot about that.

So, we need 
* a second `ranger-hive-security.xml`
* to stop the docker env 
* to switch the files 
* to start the env again

I think a second `test_all.sh` file, like you suggested, would be best. It should take care of the setup as well.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -127,6 +128,9 @@ HIVE_STANDALONE_METASTORE_JAR="standalone-metastore/target/$HIVE_STANDALONE_META
 # Calcite jar
 CALCITE_CORE_JAR_NAME="calcite-core-1.36.0.jar"
 
+# Table names
+TABLE_PERSONS="persons"
```
xBis7, line 132: Can you add here, the rest of the new tables that haven't been parameterized?

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -305,8 +309,13 @@ handleHadoopEnv() {
 handleHiveEnv() {
   abs_path=$1
   op=$2
+  hive_url_policies_enabled=$3
 
   hive_docker_path="$abs_path/$PROJECT_HIVE/packaging/target/apache-hive-3.1.3-bin/apache-hive-3.1.3-bin/compose/hive-metastore-ranger"
+  if [ "$hive_url_policies_enabled" == "true" ]; then
```
xBis7, line 315: I think we should restore the original file, when stopping the env.

trino-hms-hdfs-ranger/tests/test_hive_url_policies.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
xBis7, line 1: Call that file at the end of `test_all.sh`.

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: Done.

mladjan-gadzic:
I've run `test_all.sh` script and all the tests pass.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -339,6 +353,9 @@ handleHiveEnv() {
     echo ""
     echo "'$PROJECT_HIVE' env stopped."
     echo ""
+
+    cp "$abs_path/$CURRENT_REPO/compose/hive/conf/ranger-hive-security.xml" "$hive_docker_path/conf/ranger-hive-security.xml"
```
xBis7, line 357: There is another approach to this. Instead of copying the file locally and then copying it back, we can rename it.

`ranger-hive-security-old.xml` and then after we finish, we can remove the new one and rename the old back to its original name.

What do you think?

xBis7: Apart from this, the rest LGTM!

.gitignore
```diff
@@ -7,3 +7,4 @@ compose/trino-spark/conf/spark/spark-events
 compose/trino-spark/conf/spark/hive-jars
 
 trino-hms-hdfs-ranger/test_func.sh
+compose/hive/conf/ranger-hive-security.xml
```
xBis7, line 10: This isn't needed anymore.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -339,6 +353,9 @@ handleHiveEnv() {
     echo ""
     echo "'$PROJECT_HIVE' env stopped."
     echo ""
+
+    mv "$hive_docker_path/conf/ranger-hive-security-old.xml" "$hive_docker_path/conf/ranger-hive-security.xml" 2>/dev/null || true
```
xBis7, line 357: We need a check here. This won't work if `hive_url_policies_enabled = false` because the file would have never been renamed in the first place.

With that change, we will need the parameter in `stop_docker_env.sh` as well.

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: Sure, why not. Done.

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: That's why I've added `2>/dev/null || true` so that it does not matter if it fails.

.gitignore
mladjan-gadzic: Removed.

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: Try this in your console, you'll see that it does not fail (return value is 0).
```
mv this that 2>/dev/null || true
echo $?
```

xBis7: I didn't see that. Thanks for pointing it out.

xBis7 (approved):
LGTM!

mladjan-gadzic:
@xBis7 thanks for the review and merge!

PR #12, opened 2024-03-15
Add tests for insert into table command

mladjan-gadzic:


trino-hms-hdfs-ranger/tests/trino/3_test_trino_hdfs_hive_all.sh
```diff
@@ -11,9 +11,18 @@ echo "- INFO: Updating Ranger policies. User [postgres] will now have all access
...
+sleep 15
+
 echo ""
 echo "- INFO: Ranger policies updated."
 
+echo ""
+echo "- INFO: Create table from CSV file"
 successMsg="CREATE TABLE"
-
 retryOperationIfNeeded "$abs_path" "createTrinoTable $TRINO_TABLE $HDFS_DIR" "$successMsg" "false"
```
xBis7, line 23: Why not test `$TRINO_TABLE` as well? This is all managed by HMS. I'm wondering, which user would be performing the write.

mladjan-gadzic: The only reason is that IMO we should have multiple different tests (one where it is managed by HMS and one where it is not managed by HMS).

xBis7: > The only reason is that IMO we should have multiple different tests (one where it is managed by HMS and one where it is not managed by HMS).

I'm fine with that.

Can you try to insert into the Trino table managed by HMS? I think it's still going to be user Trino.

trino-hms-hdfs-ranger/setup/setup_for_trino_spark_testing.sh
```diff
@@ -23,3 +23,10 @@ else
   echo ""
   echo "- RESULT: HDFS test data creation failed."
 fi
+
+echo ""
+if createTrinoDir; then
+  echo "- INFO: Creation of HDFS dir for storing Trino data succeeded."
+else
+  echo "- INFO: Creation of HDFS dir for storing Trino data failed."
```
GeorgeJahad, line 31: Should this line say "ERROR" instead of "INFO"?

also should it exit if the create fails?

trino-hms-hdfs-ranger/tests/test_hive_url_policies.sh
```diff
@@ -25,7 +25,12 @@ if [ "$component" == "spark" ]; then
 
 else
   echo ""
-  # Add TRINO tests
+    echo "### TEST_1 ###"
+    ./tests/spark/hive_url_policies/1_test_no_hive_url_policies.sh "$abs_path"
```
GeorgeJahad, line 29: /tests/spark?

trino-hms-hdfs-ranger/tests/test_hive_url_policies.sh
```diff
@@ -25,7 +25,12 @@ if [ "$component" == "spark" ]; then
...
+
+    echo ""
+    echo "### TEST_2 ###"
+    ./tests/spark/hive_url_policies/2_test_create_hive_url_policies.sh "$abs_path"
```
GeorgeJahad, line 33: same

trino-hms-hdfs-ranger/tests/trino/3_test_trino_hdfs_hive_all.sh
```diff
@@ -11,9 +11,18 @@ echo "- INFO: Updating Ranger policies. User [postgres] will now have all access
...
+
+echo ""
+echo "- INFO: Create table"
```
GeorgeJahad, line 26: what does this test do that the one above doesn't?

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
```diff
@@ -15,15 +15,20 @@ echo "- INFO: Updating Ranger policies. User [postgres] will now have only [sele
...
+echo "- INFO: Rename table"
 echo "- INFO: [alter] should fail."
-
 failMsg="Permission denied: user [postgres] does not have [ALTER] privilege"
-
 retryOperationIfNeeded "$abs_path" "alterTrinoTable $TRINO_TABLE $NEW_TRINO_TABLE_NAME" "$failMsg" "true"
+
+echo ""
+echo "- INFO: Insert into table"
+echo "- INFO: [alter] should fail."
```
GeorgeJahad, line 31: should this be "[insert] should fail"?

trino-hms-hdfs-ranger/tests/trino/5_test_trino_select_alter_perm.sh
```diff
@@ -15,8 +15,14 @@ echo "- INFO: Updating Ranger policies. User [postgres] will now have [select, a
...
+echo "- INFO: Rename table"
 echo "- INFO: [alter] should now succeed."
-
 successMsg="RENAME TABLE"
-
 retryOperationIfNeeded "$abs_path" "alterTrinoTable $TRINO_TABLE $NEW_TRINO_TABLE_NAME" "$successMsg" "false"
+
+echo ""
+echo "- INFO: Insert into table"
+echo "- INFO: [alter] should succeed."
```
GeorgeJahad, line 25: insert?

GeorgeJahad:
I would like to document the dump files better.  I'm not quite sure of the best way yet, but for now let's try this:

Add a comment to the top of the dump file.  The comment should list all policies in the dump file that are different from the default policies.  Add enough detail so that someone can recreate the dump file policies by hand.

For this PR, just do the two dump files that are being modified.

Let me know if that is too hard, and we'll try something else.

trino-hms-hdfs-ranger/tests/trino/3_test_trino_hdfs_hive_all.sh
mladjan-gadzic: > Can you try to insert into the Trino table managed by HMS? I think it's still going to be user Trino.

Did that. You are correct, it is trino user.

trino-hms-hdfs-ranger/setup/setup_for_trino_spark_testing.sh
mladjan-gadzic: Fixed.

trino-hms-hdfs-ranger/tests/test_hive_url_policies.sh
mladjan-gadzic: Fixed.

trino-hms-hdfs-ranger/tests/test_hive_url_policies.sh
mladjan-gadzic: Fixed.

trino-hms-hdfs-ranger/tests/trino/3_test_trino_hdfs_hive_all.sh
mladjan-gadzic: One was creating non-managed and other managed Hive table.

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
mladjan-gadzic: It should not, because it is an alter not an insert permission.

trino-hms-hdfs-ranger/tests/trino/5_test_trino_select_alter_perm.sh
mladjan-gadzic: Answered this here https://github.com/xBis7/docker-setup-helper-scripts/pull/12#discussion_r1528218001.

mladjan-gadzic:
> I would like to document the dump files better. I'm not quite sure of the best way yet, but for now let's try this:
> 
> Add a comment to the top of the dump file. The comment should list all policies in the dump file that are different from the default policies. Add enough detail so that someone can recreate the dump file policies by hand.
> 
> For this PR, just do the two dump files that are being modified.
> 
> Let me know if that is too hard, and we'll try something else.

Did that. Please have a look at changes and let me know if it is understandable.

mladjan-gadzic:
On top of fixing tests and addressing review comments [Fix tests and address review comments](https://github.com/xBis7/docker-setup-helper-scripts/pull/12/commits/9138b5ba47859b5518d9a24e4fbc4f91a63415b6) commit introduced some quality of life changes which makes maintaining this project easier.

trino-hms-hdfs-ranger/setup/load_ranger_policies.sh
```diff
@@ -20,7 +20,7 @@ fi
 
 echo ""
 echo "Dropping the old populated ranger DB."
-if docker exec -it ranger-postgres psql -U postgres -d postgres -c "DROP DATABASE ranger"; then
+if docker exec -it ranger-postgres psql -U postgres -d postgres -c "DROP DATABASE ranger" > /dev/null; then
```
GeorgeJahad, line 23: I'd like to be able to see the error if it fails.  Can we redirect to a tmp file, and dump that file on error?

GeorgeJahad:
> Did that. Please have a look at changes and let me know if it is understandable.

Thanks for adding the ascii policies.  They help, but it is hard to review these tests if the test setup isn't documented.  For example, which users are associated with which actions?  which users are in the public group?  Are these questions answered anywhere?

Also, I think many of these policies are irrelevant, right?  for example all the polices for the hive user, can they be ignored/removed?

What documention do we need to add so someone can look at a set of changes and decide if they seem right?

I should have asked these questions earlier, when the scripts were initially being created, but I was busy with other things then and didn't notice.  But I think it would be good to document some of this stuff now.

Also, is the "/" character used to show empty groups?  That is a bit confusing, isn't it? Maybe use "-" instead, like it shows in the webui.

GeorgeJahad:
line 6 in 8_hdfs_hive_create_hive_url.sql looks off.

trino-hms-hdfs-ranger/setup/load_ranger_policies.sh
mladjan-gadzic: Done.

mladjan-gadzic:
It does not. I've rerun all tests for Trino just to be sure and none failed. Did you run scripts before tests to set up project?

mladjan-gadzic:
> this test seems to fail for me: ./tests/trino/4_test_trino_only_select_perm.sh
> 
> does it fail for you?


It does not. I've rerun all tests for Trino just to be sure and none failed. Did you run scripts before tests to set up project?

mladjan-gadzic:
> line 6 in 8_hdfs_hive_create_hive_url.sql looks off.

Fixed.

mladjan-gadzic:
> > Did that. Please have a look at changes and let me know if it is understandable.
> 
> Thanks for adding the ascii policies. They help, but it is hard to review these tests if the test setup isn't documented. For example, which users are associated with which actions? which users are in the public group? Are these questions answered anywhere?

I agree that it's hard to review - and it is only going to be harder if we add more tests. That's why I proposed to refactor this and have somekind of organization. Questions you've mentioned are not really answered anywhere - the best you can do is to read test line by line to understand what is happening.
> 
> Also, I think many of these policies are irrelevant, right? for example all the polices for the hive user, can they be ignored/removed?
> 

You are right, but the reason I added those there is that by default `{OWNER}, Hive, ...` users are added to every single policy - which is not what we want. On example if `spark` user creates a table it has `Drop` permission by default because it is owner of that table. In order to test every single permission isolated `{OWNER}` user was removed from default permissions and only `Hive` is left.
> What documention do we need to add so someone can look at a set of changes and decide if they seem right?
> 

IMO decoupling and organizing tests in a different manner might help with this. Also, documenting default policies and then using those to build upon might be helpful. So when someone sees policies like `default + Create for spark user for Hive URL policy` might be more understandable.
> I should have asked these questions earlier, when the scripts were initially being created, but I was busy with other things then and didn't notice. But I think it would be good to document some of this stuff now.
> 

Would you like to create a ticket and explain what exactly should we document? I'd try to come up with an ideas how and you can approve those or suggest something else?
> Also, is the "/" character used to show empty groups? That is a bit confusing, isn't it? Maybe use "-" instead, like it shows in the webui.

I agree. I'll use `-` instead.

trino-hms-hdfs-ranger/tests/trino/hive_url_policies/1_test_no_hive_url_policies.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Goal of this test is to verify that user postgres can't create a table managed by Hive without Write permission for Hive URL policies.
```
mladjan-gadzic, line 3: @GeorgeJahad please let me know whether this doc/explanation is helpful and whether you'd like me to do the same for other tests.

GeorgeJahad:
for the dump comments, i'd like to remove the irrelevant info.  it seems like we can remove the groups column, and the hive user rows and the public group rows.  So the comment for ranger_dumps/8_hdfs_hive_create_hive_url.sql could look like this:
```
-- Service Name     | Policy Name                                   | Users                             | Permissions
--------------------|-----------------------------------------------|-----------------------------------|-------------
-- hadoopdev        | all-path                                      | hadoop, postgres, spark           | Read, Write, Execute
-- hivedev          | all - url                                     | spark, postgres                   | write                     
-- hivedev          | default database tables column                | postgres, spark                   | *

```
if you agree, then fix it and the others that you have modified.

trino-hms-hdfs-ranger/tests/trino/hive_url_policies/1_test_no_hive_url_policies.sh
GeorgeJahad: This doesn't seem quite right either.  I would leave it here for now, but don't do it for any of the others yet.

GeorgeJahad (approved):
I'm not quite done reviewing this, but I don't want to keep you from moving forward.  If it is keeping you from working on the h4 merge, go ahead and merge it.  (I have approved it for now so you can, but may still have some comments over the next day or so and will add them to a separate jira ticket if I need to.)

mladjan-gadzic:
Merging this so that @xBis7 can continue his work. There are issues with Hive 4 and Trino tests (delete from Hive managed table and URL policies). Working on fix right now.

PR #15, opened 2024-03-28
Remove unnecessary Hive configs from Spark

xBis7:
While working on https://github.com/G-Research/gr-oss/issues/620, I did extensive testing of all the Hive security configs. I found out that Spark, just like Trino, doesn't need to be aware of any Hive security properties at all. Security must be set up only on the HiveMetastore side.


mladjan-gadzic (approved):
Thanks for this improvement. It is always great to reduce maintanance effort when possible.

xBis7:
@mladjan-gadzic Thanks for the approval. I'm merging this.

PR #16, opened 2024-03-28
gr-oss-626. Finish documenting dump files.

xBis7:


trino-hms-hdfs-ranger/ranger_dumps/3_hdfs_all.sql
```diff
@@ -22,7 +36,8 @@ SET row_security = off;
 
 CREATE FUNCTION public.getmodulesidbyname(input_val character varying) RETURNS bigint
     LANGUAGE sql
-    AS $_$ SELECT x_modules_master.id FROM x_modules_master
+    AS $_$ SELECT x_modules_master.id FROM x_modules_master
+
```
xBis7, line 40: It seems that the editor added some extra spaces here and there.

trino-hms-hdfs-ranger/ranger_dumps/1_defaults.sql
```diff
@@ -1,3 +1,19 @@
+-- Service Name     | Policy Name                                   | Group        | Users                                  | Permissions
+--------------------|-----------------------------------------------|--------------|----------------------------------------|---------------------
+-- hadoopdev        | all-path                                      | -            | hadoop                                 | Read, Write, Execute
```
mladjan-gadzic, line 3: While doing documentation for the rest of PG dumps, @GeorgeJahad asked me to keep only relevant info. Could you please do the same? [Here](https://github.com/xBis7/docker-setup-helper-scripts/pull/12#issuecomment-2008468821) is an example. This is still not in master, and that is why this was unkown to you.

mladjan-gadzic: Same applies for other PG dumps.

mladjan-gadzic: To be precise, documentation of this dump could be as follow:
```
-- Service Name     | Policy Name                                   | Users                                  | Permissions
--------------------|-----------------------------------------------|----------------------------------------|---------------------
-- hadoopdev        | all-path                                      | hadoop                                 | Read, Write, Execute
```

xBis7: ooh, I see and you had also done the same

https://github.com/xBis7/docker-setup-helper-scripts/pull/12/commits/df286720dae3f5ebc926e0fb3da75b0e1c39f3d1

But the info isn't irrelevant.

Here public group means that almost everyone can create.

```
-- hivedev | default database tables columns | public | - | Create
```

but if I remove the group and make it like this, anyone reading it will think that no one has permissions

```
-- hivedev | default database tables columns | - | Create
```

which is the opposite of the truth.

And here we used to have issues with the {OWNER} tag and we need to show it existing here in order to show in the next dumps that we removed it. 

Also, we have a database test suite. All these policies are relevant to these tests. If we change any of these, the tests will fail.

```
-- hivedev          | all - database, table, column                 | -            | hive, {OWNER}                          | *
-- hivedev          | all - database, table                         | -            | hive, {OWNER}                          | *
-- hivedev          | all - database                                | -            | hive, {OWNER}                          | *
```

xBis7: BTW, we removed the public group from the second dump, because when trying to create a table with Trino, we expected the operation to fail entirely but it was actually half-way succeeding.

Because the process of create is
* Create tmp (**success**, public group has create permission)
* Alter, rename tmp to expected name (**failure**, public group has no permission)

mladjan-gadzic: > But the info isn't irrelevant.

I understand that. What I think that @GeorgeJahad's  intention was is to document only the most important ones. If it makes sense from test perspective to have for example `public` group documented, let's leave it there.

However, for example, in `3_hdfs_all.sql` dump, maybe we can remove rows with `hive` user, since it just pollutes doc (because hive user needs to have those permissions anyway and we are not changing those afterward). In that sense it is more important to document `hadoopdev`. I know that it is different from `1_default.sql` because `{OWNER}` user is not there right now, but that's my understanding of how @GeorgeJahad would like to see that. I've also documented everything like you did, and then cut down to only "the most important" ones.

Nevertheless, if you think those are necessary, it is okay to leave them, I am not against it. I do understand those permissions quite well. So no objections from my side.

xBis7: The only policy we modify for the database tests, is this one

```
-- hivedev          | all - database, table, column                 | -            | hive, spark, postgres                  | *
```

but TBH I don't if it makes a difference that we haven't added users `spark` and `postgres` in 

```
-- hivedev          | all - database                                | -            | hive                                   | *
```

I think it would give us the same result. Also, I don't remember in which of all these policies, the `{OWNER}` tag was creating the issues.

There is a lot of overlap between the database and table policies and I don't know which ones to cleanup. We can delete them all and start fresh with just the minimum, only the ones we need for the tests. Documentation will be simpler as well. 

Let's leave this PR open and discuss it further next week that @GeorgeJahad will be back.

mladjan-gadzic: >There is a lot of overlap between the database and table policies and I don't know which ones to cleanup. We can delete them all and start fresh with just the minimum, only the ones we need for the tests. Documentation will be simpler as well.

During discussion with @GeorgeJahad, this came up as an idea.

> Let's leave this PR open and discuss it further next week that @GeorgeJahad will be back.

I agree. There should not be any large conflicts (if any).

xBis7:
In PR https://github.com/xBis7/docker-setup-helper-scripts/pull/24, I recreated all the DB dump files and also documented them along the way. This PR is not needed anymore. I'm closing it.

PR #18, opened 2024-03-29
gr-oss-625. Improve retryOperationIfNeeded() output

xBis7:
https://github.com/G-Research/gr-oss/issues/625

This is improving the output of the `retryOperationIfNeeded` method. Each method execution is redirected to a file using the `tee` command. The `tee` command gets the output and prints it to the stdout while also writing it to a file.

I've added an echo with the current command
```
Command being run: -- '...'
```

and I've also wrapped the stdout like this, so that it's easier to spot
```
### Stdout- {

...

} -Stdout ###
```

Example output

```
Command being run: -- 'spark.read.text("hdfs://namenode:8020/test").write.option("path", "hdfs://namenode/opt/hive/data").mode("overwrite").format("csv").saveAsTable("poc_db.spark_test_table")'


### Stdout- {

24/03/29 16:16:37 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240329161637-0005).
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 3.3.2
      /_/
         
Using Scala version 2.12.15 (OpenJDK 64-Bit Server VM, Java 11.0.22)
Type in expressions to have them evaluated.
Type :help for more information.

scala> Hive Session ID = 582b5268-b4f7-4c27-8618-6eed1dbd59a1
24/03/29 16:16:44 WARN HiveExternalCatalog: Couldn't find corresponding Hive SerDe for data source provider csv. Persisting data source table `poc_db`.`spark_test_table` into Hive metastore in Spark SQL specific format, which is NOT compatible with Hive.
org.apache.spark.sql.AnalysisException: org.apache.hadoop.hive.ql.metadata.HiveException: MetaException(message:Permission denied: user [spark] does not have [CREATE] privilege on [poc_db/spark_test_table])
  at org.apache.spark.sql.hive.HiveExternalCatalog.withClient(HiveExternalCatalog.scala:110)
...
...
...
...
...
...
...
...
...
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
  at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
  at java.base/java.lang.reflect.Method.invoke(Unknown Source)
  at org.apache.hadoop.hive.metastore.RetryingMetaStoreClient.invoke(RetryingMetaStoreClient.java:212)
  at com.sun.proxy.$Proxy40.createTable(Unknown Source)
  at org.apache.hadoop.hive.ql.metadata.Hive.createTable(Hive.java:928)
  ... 93 more

scala> :quit


} -Stdout ###
```

I've run all tests for both Spark and Trino multiple times to make sure nothing is broken.

The tricky part has been printing the command. If we add the print inside every method that runs the command, then the print will also get stored in the tmp file that we are checking for the expected output message. Some times the expected output is part of the command and we end up with unexpected failures.

### DIFF
--- a/trino-hms-hdfs-ranger/testlib.sh
+++ b/trino-hms-hdfs-ranger/testlib.sh
@@ -725,10 +725,12 @@ createTrinoTable() {
   hdfs_dir_name=$2
   schema_name=$3
 
+  c="create table hive.$schema_name.$table_name (column1 varchar,column2 varchar) with (external_location = 'hdfs://namenode:8020/$hdfs_dir_name',format = 'CSV');"
+
   if [ "$PRINT_CMD" == "true" ]; then
-    printCmdString "create table hive.$schema_name.$table_name (column1 varchar,column2 varchar) with (external_location = 'hdfs://namenode:8020/$hdfs_dir_name',format = 'CSV');"
+    printCmdString "$c"
   else
-    docker exec -it "$TRINO_HOSTNAME" trino --execute="create table hive.$schema_name.$table_name (column1 varchar,column2 varchar) with (external_location = 'hdfs://namenode:8020/$hdfs_dir_name',format = 'CSV');"
+    docker exec -it "$TRINO_HOSTNAME" trino --execute="$c"
   fi
 }
 
@@ -736,10 +738,12 @@ selectDataFromTrinoTable() {
   table_name=$1
   schema_name=$2
 
+  c="select * from hive.$schema_name.$table_name;"
+
   if [ "$PRINT_CMD" == "true" ]; then
-    printCmdString "select * from hive.$schema_name.$table_name;"
+    printCmdString "$c"
   else
-    docker exec -it "$TRINO_HOSTNAME" trino --execute="select * from hive.$schema_name.$table_name;"
+    docker exec -it "$TRINO_HOSTNAME" trino --execute="$c"
   fi
 }
 
@@ -748,10 +752,12 @@ alterTrinoTable() {
   new_table_name=$2
   schema_name=$3
 
+  c="alter table hive.$schema_name.$old_table_name rename to $new_table_name;"
+
   if [ "$PRINT_CMD" == "true" ]; then
-    printCmdString "alter table hive.$schema_name.$old_table_name rename to $new_table_name;"
+    printCmdString "$c"
   else
-    docker exec -it "$TRINO_HOSTNAME" trino --execute="alter table hive.$schema_name.$old_table_name rename to $new_table_name;"
+    docker exec -it "$TRINO_HOSTNAME" trino --execute="$c"
   fi
 }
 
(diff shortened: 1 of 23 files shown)


mladjan-gadzic (approved):
Handy improvement, thanks! I've tested everything - no failures.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -565,142 +570,222 @@ updateProjectRepo() {
...
   hdfs_dir_name=$2
   db_name=$3
+  print=$4
 
-  if [ "$db_name" == "" ]; then
-    db_name="default"
+  if [ "$print" == "true" ]; then
+    printCmdString "spark.read.text(\"hdfs://namenode:8020/$hdfs_dir_name\").write.option(\"path\", \"hdfs://namenode/opt/hive/data\").mode(\"overwrite\").format(\"csv\").saveAsTable(\"$db_name.$table_name\")"
+  else
+    docker exec -it "$SPARK_MASTER_HOSTNAME" bash -c "echo \"spark.read.text(\\\"hdfs://namenode:8020/$hdfs_dir_name\\\").write.option(\\\"path\\\", \\\"hdfs://namenode/opt/hive/data\\\").mode(\\\"overwrite\\\").format(\\\"csv\\\").saveAsTable(\\\"$db_name.$table_name\\\")\" | bin/spark-shell"
```
GeorgeJahad, line 639: I don't like how we have two differing copies of the command in each of these subcommands.  I would much prefer a structure like this:

```
cmd = xyz
if print == true
   printCmdString $cmd
else $cmd
```

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -839,13 +924,45 @@ retryOperationIfNeeded() {
...
+    # * The expected output is just the phrase 'CREATE SCHEMA'
+    # * The command to create a schema, starts with the phrase 'CREATE SCHEMA ...'
+    # * We do a 'grep' on the tmp file to check that the output contains 'CREATE SCHEMA'
+    # * The command failed and the output is some error
+    # * 'CREATE SCHEMA' exists in the command, which is present in the tmp file.
+    # * 'grep' succeeds, exit status will be 0 regardless of the output.
+    #
+    # To fix that, each method will either print the command or execute it.
+    # We will call each method twice. Once with the print parameter and once without.
+    $cmd "true"
```
GeorgeJahad, line 942: It is not great to have the last parameter have a special meaning.  Now people have to remember when they add a parameter to make it the second to last.  It is not very elegant.  Can you do something else?  Maybe use a global variable?

xBis7: Done.

trino-hms-hdfs-ranger/testlib.sh
xBis7: That was my first approach but it's a bit tricky to properly handle the quotation marks. I'll try to find a way.

xBis7:
@mladjan-gadzic @GeorgeJahad Can you take another look? I've made all the changes and I've also run all tests for Spark and Trino and they all pass.

trino-hms-hdfs-ranger/testlib.sh
xBis7: Fixed.

GeorgeJahad (approved):
Thanks for the fixes.  LGTM.  Merging so that Mladjan can integrate his fixes.

PR #20, opened 2024-04-02
gr-oss-584. Remove jars with build different from current one

mladjan-gadzic:
[gr-oss-584](https://github.com/G-Research/gr-oss/issues/584)

Right now, Spark throws `java.lang.NoSuchFieldError: SPLIT_UPDATE` exception after changing Hive 3 and Hive 4 builds multiple times, because older jars are not properly removed.

This PR fixes issue when there are multiple Hive jars from different builds. It aims to delete older one and leave one from current build.

### DIFF
--- a/trino-hms-hdfs-ranger/testlib.sh
+++ b/trino-hms-hdfs-ranger/testlib.sh
@@ -193,13 +193,23 @@ setupSparkJarsIfNeeded() {
   dir_base_path="$abs_path/$CURRENT_REPO/compose/spark/conf"
   jars_dir_name="hive-jars"
   jars_dir_path="$dir_base_path/$jars_dir_name"
-  file_name_regex="hive-*-$HIVE_BUILD*"
+  hive_jar_regex_prefix="hive-*"
+  # Flag to track if any file does not contain $HIVE_BUILD
+  delete_files=false
 
-  # Check whether compatible jar files already exist. If not cleanup jars dir.
-  if ! find "$jars_dir_path" -type f -name "$file_name_regex" -print -quit | grep -q .; then
-    echo "File matching regex '$file_name_regex' does not exist in '$jars_dir_name'."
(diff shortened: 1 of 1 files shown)


trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -193,13 +193,23 @@ setupSparkJarsIfNeeded() {
   dir_base_path="$abs_path/$CURRENT_REPO/compose/spark/conf"
   jars_dir_name="hive-jars"
   jars_dir_path="$dir_base_path/$jars_dir_name"
-  file_name_regex="hive-*-$HIVE_BUILD*"
+  hive_jar_regex_prefix="hive-*"
+  # Flag to track if any file does not contain $HIVE_BUILD
+  delete_files=false
+
+  for file in $jars_dir_path/$hive_jar_regex_prefix; do
```
xBis7, line 200: Does this have to contain the file prefix?

Shouldn't it be `$jars_dir_path/*` to make sure it lists all files?

xBis7: And we might also want to check whether the file is an actual file or a dir.

xBis7: I was wondering whether it would be better to search for all files under the directory instead of the ones matching a pattern. 

In that case, where we test all files, we should also test that the current file at hand isn't a directory.

But the current approach is sufficient for the existing implementation.

xBis7 (approved):

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: Oh, I get it. Anyways we can make such a change quickly in a future if necessary.

mladjan-gadzic:
@xBis7 thanks for the review!

PR #21, opened 2024-04-03
gr-oss-624. Test Results not being correctly checked.

mladjan-gadzic:
[gr-oss-624](https://github.com/G-Research/gr-oss/issues/624)

Trino provides correct exit code which is used in combination with expected messages.

For Spark, the idea is to write tests in Scala, set exit code depending on the test and use those files interactively with Spark shell. Updated code that works for Trino works for Spark as well with no additional changes.

Tests for Hive3/Trino/Spark were run without failures.
Tests for Hive4/Trino were run with the same failures as described in [gr-oss-643](https://github.com/G-Research/gr-oss/issues/643).


xBis7 (commented):
Right now, there are 28 new files that are pretty similar to each other. I think that we should create 1 or 2 parameterized files. 

I was able to do it locally. Check this diff: https://github.com/xBis7/docker-setup-helper-scripts/commit/2d66a1b041f7f0f2fe87ffbefe14c0eaeb56d155

I've tested it and here is the output from parameterizing `3_2_test.scala`

```
- INFO: Counter=0

Command being run: -- 'Running Spark test file: 3_2_test.scala'


### Stdout- {

24/04/04 18:06:56 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240404180657-0002).
Str: create table animals (id int, name string) using parquet partitioned by (name)
Hive Session ID = 2fcb0d9c-d420-4b98-b772-d5409227aa9a
Test passed

} -Stdout ###


- Expected Msg: 

- RESULT -> SUCCESS: Operation succeeded as expected.
```

I've added parameters for the sql command and the expected message but we can add other parameters as well depending on our needs.

That way, we can also print the command.

This will be printing the actual command, like before.
```
Command being run: -- 'Running Spark test file: 3_2_test.scala'
```

GeorgeJahad:
>  I think that we should create 1 or 2 parameterized files.

This is a good idea, (in fact you don't need any files in git at all.  you could just create them all at runtime.)  I don't want to spend too much time on it right now.

If you think it is better, and think you can do it in less than a day, go ahead and do it.  Otherwise, just create another ticket for it and let's get this merged now.

Either way, I do want to see the command in the output, so if you don't do this, then just cat the file to stdout when you do the docker copy.

trino-hms-hdfs-ranger/tests/spark/database/6_test.scala
```diff
@@ -0,0 +1,14 @@
+try {
+  spark.sql("drop database if exists poc_db")
+  println("Test failed. Database dropped successfully.")
+  sys.exit(1)
+} catch {
+  // This is error message for Spark 3.5.0: '[SCHEMA_NOT_EMPTY] Cannot drop a schema'
```
GeorgeJahad, line 6: instead of comments, maybe create variables and just use the 3.3.2 variable for now.

GeorgeJahad:
For some reason, when I run ./tests/trino/schema/7_drop_schema_cascade.sh I get this error:
```
Query 20240404_220307_00012_5tkev failed: CASCADE is not yet supported for DROP SCHEMA
```

Any ideas?

And I'm still getting the error reported here: https://github.com/G-Research/gr-oss/issues/623

GeorgeJahad (approved):

xBis7:
> For some reason, when I run ./tests/trino/schema/7_drop_schema_cascade.sh I get this error:
> 
> Query 20240404_220307_00012_5tkev failed: CASCADE is not yet supported for DROP SCHEMA
> Any ideas?

Are you creating the Schema manually and then running test number 7?

I just ran all the Schema tests `./tests/trino/schema/test_schema.sh /path/to/projects`

and it worked. This is the output for test number 7

```
Configuring project for Hive 3.


- INFO: User [postgres] has the correct HDFS and Hive permissions to drop a DB.
- INFO: Dropping a DB that's not empty with CASCADE, should succeed.

- INFO: Some wait time might be needed for the policy update to get picked up. Retry a few times if needed.

- INFO: Counter=0

Command being run: -- 'DROP SCHEMA hive.poc_db CASCADE;'


### Stdout- {

DROP SCHEMA

} -Stdout ###


- Expected Msg: DROP SCHEMA

- RESULT -> SUCCESS: Operation succeeded as expected.
---------------------------------------------------
```

But your output indicates that CASCADE isn't supported. The first thing that comes to my mind, is that you are using an older version of Trino. We run Trino from the official latest docker image. Could it be cached and latest be pointing to an image from 1 year ago or something like that?

> And I'm still getting the error reported here: https://github.com/G-Research/gr-oss/issues/623

As @mladjan-gadzic mentioned, the tests depend on each other and that's the reason their names are numbered. We have a script for setting up the HDFS resources that will allow you to skip running the previous scripts.

I'm using it all the time. That's the command

```
./setup/setup_for_trino_spark_testing.sh /path/to/projects
```

If you run this, then you will be able to run this without errors

```
./tests/trino/4_test_trino_only_select_perm.sh
```

Just an idea, one of us might be getting an error that the others don't get, and it will probably be due to the env setup. We all have the same setup inside the docker containers but not outside of them. We could be doing what spark is doing with the `build` directory. We could download all the tools we need and the specific versions that we know work and use them for setting up the env. When I'm saying tools, I mean Java8, Java17, docker-compose etc.

mladjan-gadzic:
@xBis7 I really like the idea of parametrizing and using only few templates for tests. I will try it out and let you know how it goes. Thanks for the suggestion and diff.

@GeorgeJahad I've re-run tests and I do not see failures. Please try to follow steps that @xBis7 has explained. After getting and building all the projects, running `tests/test_all.sh $ABS_PATH trino true` should be enough.

mladjan-gadzic:
I was able to reduce number of scala files to five. There is one special case `4_1_test.scala` which should be left as is. Other four can be reduced to two `test` and `test_external_table_creation`, but I find that it might become hard to read. That is the reason I separated testing without and for an exception.

@xBis7 please try to run tests on your machine. If those are failing, please let me know. I've created `base64encode` function in `testlib.sh` specifically for this occation so that it can be modified to include OS differences and act on those accordingly.

Currently I am parametrizing table names in commands.

Example of command print 
```
Command being run: -- 'Running Spark test: [test_for_exception.scala] with arguments sql_arg: [create table persons (id int, name string)] and msg_arg: [Permission denied: user [spark] does not have [WRITE] privilege on [[hdfs://namenode/opt/hive/data/persons, hdfs://namenode/opt/hive/data/persons/]]]'
```

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -617,17 +624,6 @@ createHdfsFile() {
...
-  sql="$*"
-
-  if [ "$PRINT_CMD" == "true" ]; then
-    printCmdString "$sql"
-  else
-    docker exec -it "$SPARK_MASTER_HOSTNAME" bash -c "echo \"$sql\" | bin/spark-shell"
-  fi
-}
-
 createSparkTable() {
```
xBis7, line 627: Can we remove this as well? Or is it just the methods that were using `spark.sql()`? 

I like the idea of reducing and simplifying the code-base. Nice!

trino-hms-hdfs-ranger/tests/spark/database/7_drop_database_cascade.sh
```diff
@@ -7,9 +7,9 @@ set -e
...
-echo "- INFO: User [spark] has the correct HDFS and Hive permissions to drop a DB."
-echo "- INFO: Dropping a DB that's not empty with CASCADE, should succeed."
-
-successMsg="res0: org.apache.spark.sql.DataFrame = []"
-
-retryOperationIfNeeded "$abs_path" "dropDatabaseWithSpark $EXTERNAL_DB true" "$successMsg" "false"
+echo "- INFO: Reusing policies."
+echo "- INFO: Drop database."
+echo "- INFO: User [spark] should be able to drop non-empty database with 'cascade'."
+cpSparkTest $(pwd)/$SPARK_TEST_PATH/$SPARK_TEST_EXTERNAL_TABLE_CREATION_NO_EXCEPTION_FILENAME
```
mladjan-gadzic, line 13: Fixing this and command below

trino-hms-hdfs-ranger/testlib.sh
xBis7: Can you edit the file to create a table just like the other tests and remove the method?

mladjan-gadzic: Sure. Done.

trino-hms-hdfs-ranger/tests/spark/database/7_drop_database_cascade.sh
mladjan-gadzic: Fixed.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -81,6 +81,13 @@ TRINO_HOSTNAME="trino-coordinator-1"
 
 SPARK_MASTER_HOSTNAME="spark-master-1"
 SPARK_WORKER1_HOSTNAME="spark-worker-1"
+SPARK_TEST_FILENAME="test.scala"
```
xBis7, line 84: Nit: 

They have been added in the hostname group. Can you separate them with a line and add a comment?

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -940,24 +845,31 @@ retryOperationIfNeeded() {
...
+    exp_exit_code=0
   fi
 
   echo ""
   echo "- INFO: Some wait time might be needed for the policy update to get picked up. Retry a few times if needed."
   echo ""
 
   counter=0
+  while [[ "$counter" -le 10 ]]; do
+    if [ "$counter" -eq 10 ]; then
```
xBis7, line 865: This used to be like this 

```
while [ counter < 9 ]; do

  counter increase

  if [ counter == 9]; then
  
  fi
done 
```

but now it's

```
while [ counter < 10 ]; do
  if [ counter == 10]; then
  
  fi
  counter increase
done 
```

The if statement will never execute.
* during loop, counter = 9
* counter increase, counter = 10
* exit loop

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -984,82 +896,62 @@ retryOperationIfNeeded() {
     echo "### Stdout- {"
     echo ""
 
+    sleep 3
+
     # Use the 'tee' command, which prints the output in the stdout
     # and writes it in a file at the same time.
     $cmd 2>&1 | tee "$abs_path/$CURRENT_REPO/$TMP_FILE"
+    exit_code=${PIPESTATUS[0]}
```
xBis7, line 904: Because we are using the `tee` command here, the exit status will always be 0. The `tee` has to fail for the exit status to be something else.

Here is a simple example
```shell
> ls -lah | grep 'bin' 2>&1 | tee /tmp/test.txt
drwxrwxr-x  2 xbis xbis 4,0K Απρ   4 14:29 bin
drwxrwxr-x  2 xbis xbis 4,0K Μαρ  29 21:18 binder
-rw-rw-r--  1 xbis xbis  23K Μαρ  29 21:18 LICENSE-binary
drwxrwxr-x  2 xbis xbis 4,0K Μαρ  29 21:18 licenses-binary
-rw-rw-r--  1 xbis xbis  57K Μαρ  29 21:18 NOTICE-binary
drwxrwxr-x  2 xbis xbis 4,0K Μαρ  29 21:18 sbin
> echo $?
0
> ls -lah | grep 'zbin' 2>&1 | tee /tmp/test.txt
> echo $?
0
> ls -lah | grep 'zbin' 2>&1
> echo $?                   
1
```

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -1071,3 +963,31 @@ waitForPoliciesUpdate() {
...
+  if [ "$PRINT_CMD" == "true" ]; then
+      printCmdString "$message"
+  else
+    docker exec -it "$SPARK_MASTER_HOSTNAME" bash -c "bin/spark-shell --conf spark.app.sql=\"$sql_arg\" --conf spark.app.msg=\"$msg_arg\" -I test.scala"
+  fi
+}
+
+# This function is created to allow for substitutes or changes if base64 works differently on different systems.
+# We can achieve the correct variable substitution and space handling by encoding and decoding the string.
+# -w 0, makes sure that the line length is ignored.
```
xBis7, line 988: Do you mean that this will be extended to handle things differently according to each platform?

trino-hms-hdfs-ranger/tests/spark/2_test_spark_hdfs_perms_no_hive_perms.sh
```diff
@@ -9,21 +9,12 @@ abs_path=$1
...
-
-echo ""
-echo "- INFO: Ranger policies updated."
-
-# Failure due to lack of Hive metastore permissions.
-
-failMsg="Permission denied: user [spark] does not have [CREATE] privilege on [default/$SPARK_TABLE]"
-
-retryOperationIfNeeded "$abs_path" "createSparkTable $SPARK_TABLE $HDFS_DIR $DEFAULT_DB" "$failMsg" "true"
+waitForPoliciesUpdate
```
xBis7, line 13: I can see that this added to every test. Does this mean that we will have to wait 30 secs before every test runs? We have updated the policy refresh threshold to 10 secs. I think 15-20 secs would be enough.

GeorgeJahad:
> the tests depend on each other and that's the reason their names are numbered. 

I'm just running run_all.sh

GeorgeJahad:
> Could it be cached and latest be pointing to an image from 1 year ago or something like that?
that was the problem, thanks!  It also fixed this issue:
https://github.com/G-Research/gr-oss/issues/623

GeorgeJahad:
These latest changes look good to me, but am I getting this error now:
```
### TEST_DATABASE ###

Configuring project for Hive 3.


Configuring project for Hive 3.

- INFO: Reusing policies.
- INFO: To create an external Database and store it in HDFS, using Spark,
- INFO: all you need is HDFS perms. No Hive perms are needed.
- INFO: User [spark] doesn't have HDFS or Hive permissions.
- INFO: Create database.
- INFO: User [spark] shouldn't be able to create database.

- INFO: Some wait time might be needed for the policy update to get picked up. Retry a few times if needed.

- INFO: Counter=0

Command being run: -- 'Running Spark test: [test_for_exception.scala] with arguments sql_arg: [create database poc_db location 'hdfs://namenode/opt/hive] and msg_arg: [/data/poc_db/external/poc_db.db']'


### Stdout- {

24/04/05 18:21:35 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240405182137-0001).
Test failed. Message: 
Syntax error at or near '''(line 1, pos 32)

== SQL ==
create database poc_db location 'hdfs://namenode/opt/hive
--------------------------------^^^


} -Stdout ###

```

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: `"$counter" -le 10` means `counter <= 10`. I've tested this out, it executes properly.

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: `PIPESTATUS` is an array variable in Bash that stores exit codes for most recently executed pipeline. Pipeline is set of most recently executed commands. For this example `ls -lah | grep 'bin' 2>&1 | tee /tmp/test.txt` there are 3 commands. `PIPESTATUS` will have 3 elements where first would be corresponding to exit code of `ls -lah` and last would be for `tee /tmp/test.txt`. I've tested this out, it works the way it should.

trino-hms-hdfs-ranger/testlib.sh
mladjan-gadzic: Yes. For example, you can run this code. If it does not work for you, we should extend this function to suit your needs with `-w` parameter.

trino-hms-hdfs-ranger/testlib.sh
xBis7: aah yes, you are right.

trino-hms-hdfs-ranger/testlib.sh
xBis7: ok, thanks for the explanation. I didn't realize it was an array.

trino-hms-hdfs-ranger/testlib.sh
xBis7: Thanks!

mladjan-gadzic:
> Reusing policies.

I've re-run this test with success.

```
Configuring project for Hive 3.

- INFO: Reusing policies.
- INFO: To create an external Database and store it in HDFS, using Spark,
- INFO: all you need is HDFS perms. No Hive perms are needed.
- INFO: User [spark] doesn't have HDFS or Hive permissions.
- INFO: Create database.
- INFO: User [spark] shouldn't be able to create database.
                                             Successfully copied 2.05kB to spark-master-1:/opt/spark/test.scala

- INFO: Some wait time might be needed for the policy update to get picked up. Retry a few times if needed.

- INFO: Counter=0

Command being run: -- 'Running Spark test: [test_for_exception.scala] with arguments sql_arg: [create database poc_db location 'hdfs://namenode/opt/hive/data/poc_db/external/poc_db.db'] and msg_arg: [Permission denied: user [spark] does not have [ALL] privilege on [hdfs://namenode/opt/hive/data/poc_db/external/poc_db.db]]'


### Stdout- {

24/04/08 08:43:30 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240408084331-0000).
Hive Session ID = b1643b4c-9fe2-4883-a41c-20ebf81a0953
Test passed

} -Stdout ###


- Expected Msg: Test passed

- RESULT -> SUCCESS: Operation succeeded as expected.
---------------------------------------------------
```

@xBis7 would you mind running this? It seems that issue is due to missing `-w` from base64 encoding. I am not 100% certain, but that is my suspecion seeing that `'` is missing from @GeorgeJahad's output.

xBis7:
@mladjan-gadzic I've run all the tests locally. I get the same output as @GeorgeJahad.

```
- Unexpected exit code. Expected: 0. Actual: 2.
- INFO: Counter=9

Command being run: -- 'Running Spark test: [test_for_exception.scala] with arguments sql_arg: [create database poc_db location 'hdfs://namenode/opt/hive] and msg_arg: [/data/poc_db/external/poc_db.db']'


### Stdout- {

24/04/08 08:54:31 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240408085431-0010).
Test failed. Message: 
Syntax error at or near '''(line 1, pos 32)

== SQL ==
create database poc_db location 'hdfs://namenode/opt/hive
--------------------------------^^^


} -Stdout ###


- Unexpected exit code. Expected: 0. Actual: 2.

- RESULT -> FAILURE: Operation exceeded the retry limit. Exiting...
---------------------------------------------------

```

When I added the `-w 0` option and reran the tests, they all passed.

mladjan-gadzic:
> @mladjan-gadzic I've run all the tests locally. I get the same output as @GeorgeJahad.
> 
> ```
> - Unexpected exit code. Expected: 0. Actual: 2.
> - INFO: Counter=9
> 
> Command being run: -- 'Running Spark test: [test_for_exception.scala] with arguments sql_arg: [create database poc_db location 'hdfs://namenode/opt/hive] and msg_arg: [/data/poc_db/external/poc_db.db']'
> 
> 
> ### Stdout- {
> 
> Setting default log level to "WARN".
> To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
> 24/04/08 08:54:31 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
> Spark context Web UI available at http://spark-master:4040
> Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240408085431-0010).
> Spark session available as 'spark'.
> Test failed. Message: 
> Syntax error at or near '''(line 1, pos 32)
> 
> == SQL ==
> create database poc_db location 'hdfs://namenode/opt/hive
> --------------------------------^^^
> 
> 
> } -Stdout ###
> 
> 
> - Unexpected exit code. Expected: 0. Actual: 2.
> 
> - RESULT -> FAILURE: Operation exceeded the retry limit. Exiting...
> ---------------------------------------------------
> ```
> 
> When I added the `-w 0` option and reran the tests, they all passed.

Thanks! I've added changes to address this.

xBis7 (approved):
@mladjan-gadzic Thanks for the fix. I can verify that all tests are passing.

mladjan-gadzic:
> @mladjan-gadzic Thanks for the fix. I can verify that all tests are passing.

Great, thank you! I am going to merge it.

PR #24, opened 2024-04-16
Use Ranger 2.4.0 and run Hadoop from the official image 3.3.6

xBis7:
The PR contains the following changes

* Ranger is now running from 2.4.0 instead of the latest version
* new Ranger DB dump files
* Hadoop has been moved under the repo and is running from the official image

All tests are passing.


mladjan-gadzic (commented):
Thanks @xBis7 for this great improvement and effort! I'll leave some minor comments inline.

compose/hadoop/conf/ranger-hdfs-audit.xml
```diff
@@ -0,0 +1,216 @@
...
+	</property>	
+
+	<property>
+		<name>xasecure.audit.hdfs.async.max.flush.interval.ms</name>
+		<value>30000</value>
+	</property>	
+
+	<property>
+		<name>xasecure.audit.hdfs.config.encoding</name>
+		<value></value>
```
mladjan-gadzic, line 48: This looks like it has no value. Is it needed then?

compose/spark/docker/docker-compose.yml
```diff
@@ -16,13 +16,13 @@ services:
...
       - ../conf/hive-jars:/opt/spark/hive-jars
       - ../conf/spark-events:/opt/spark/spark-events
-      - ../conf/spark-defaults.conf:/opt/spark/conf/spark-defaults.conf:ro
-      - ../conf/hive-site.xml:/opt/spark/conf/hive-site.xml:ro
-      - ../conf/ranger-hive-audit.xml:/opt/spark/conf/ranger-hive-audit.xml:ro
-      - ../conf/ranger-hive-security.xml:/opt/spark/conf/ranger-hive-security.xml:ro
-      - ../conf/ranger-policymgr-ssl.xml:/opt/spark/conf/ranger-policymgr-ssl.xml:ro
-      - ../../common/core-site.xml:/opt/spark/conf/core-site.xml:ro
-      - ../../common/hdfs-site.xml:/opt/spark/conf/hdfs-site.xml:ro
+      - ../conf/spark-defaults.conf:/opt/spark/conf/spark-defaults.conf
```
mladjan-gadzic, line 19: Just curious, why `:ro` needed to be removed?

xBis7: Because we are mounting the files, the `read-only` would persist on the local environment as well and it wouldn't let me clean the files.

And in another branch where Spark is run from the source code, the `ro` makes the build fail because mvn doesn't have the permission to cleanup the target dir that contains the files. To workaround it, I had to delete the files manually from the terminal with `sudo`.

trino-hms-hdfs-ranger/ranger_dumps/6_hive_defaultdb_select_alter.sql
mladjan-gadzic, line 1: Why trino user is added to hivedev/default database tables columns?

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -286,20 +293,44 @@ setupSparkJarsIfNeeded() {
   cpJarIfNotExist "$jars_dir_path" "$hive_standalone_metastore_jar_path" "$HIVE_STANDALONE_METASTORE_JAR_NAME"
 }
 
+setupRangerJarsIfNeeded() {
```
mladjan-gadzic, line 296: I think there is similar function for Spark jars. Is there a way to create a single method instead? For example, it would accept abs_path, dir_base_path, jars_dir_name, project and array of jars. For this to work two global arrays need to be created for Spark and Ranger. Also, instead of creating jar path and copying we would iterate through passed array and do the same iteratively. Cons of this are that there would be 5 parameters to be passed in function call.

mladjan-gadzic: Consider this an optional if it is too big of a trouble.

trino-hms-hdfs-ranger/ranger_dumps/6_hive_defaultdb_select_alter.sql
xBis7: This must have been a leftover from trying to resolve some errors. I verified that we don't need trino user and removed it.

trino-hms-hdfs-ranger/testlib.sh
xBis7: The copy method is like this
```
cpJarIfNotExist $jars_dir_path $external_jar_path $jar_name
```

The `jars_dir_path` is common but for every jar you need both the `external_jar_path` and the `jar_name`.

An array of jars wouldn't help. It has to be something like a HashMap, so that any entry has both the jar path and the name. I don't know if that's doable in a shell script.

Also, we would be moving all the jar listing to the outer method that will call this new method.

xBis7:
I converted this PR into a draft because I think that the messages that we check for our current tests are wrong.

All of our tests that were checking for HDFS access, they were getting an exception like this one
```
org.apache.hadoop.security.AccessControlException
```
and the error message was also containing the Hadoop ACLs. This means that it was using Hadoop ACLs instead of Ranger. The exception should have been 
```
org.apache.ranger.authorization.hadoop.exceptions.RangerAccessControlException
```
and there should have been no ACLs in the error message.

Here is an example of a correct exception that proves that Ranger denied the access

```
Caused by: org.apache.ranger.authorization.hadoop.exceptions.RangerAccessControlException: Permission denied: user=spark, access=EXECUTE, inode="/test"
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkRangerPermission(RangerHdfsAuthorizer.java:502)
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkPermissionWithContext(RangerHdfsAuthorizer.java:232)
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:240)
	... 19 more
```

GeorgeJahad (approved):
Christos, after our discussion in standup, these changes look good to me.

xBis7:
I'm still getting this error.

```
Command being run: -- 'insert into hive.default.trino_test_table values ('5', 'cat');'


### Stdout- {

Query 20240417_164840_00023_acs7j failed: Create temporary directory for hdfs://namenode:8020/test failed: Permission denied: user=trino, access=WRITE, inode="/":hadoop:supergroup:drwxr-xr-x
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.check(FSPermissionChecker.java:506)
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:346)
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkDefaultEnforcer(RangerHdfsAuthorizer.java:635)
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkRangerPermission(RangerHdfsAuthorizer.java:395)
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkPermissionWithContext(RangerHdfsAuthorizer.java:232)
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:240)
	at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPermission(FSDirectory.java:1943)
	at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPermission(FSDirectory.java:1927)
	at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkAncestorAccess(FSDirectory.java:1886)
	at org.apache.hadoop.hdfs.server.namenode.FSDirMkdirOp.mkdirs(FSDirMkdirOp.java:60)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.mkdirs(FSNamesystem.java:3441)
	at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.mkdirs(NameNodeRpcServer.java:1167)
	at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.mkdirs(ClientNamenodeProtocolServerSideTranslatorPB.java:742)
	at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
	at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:621)
	at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:589)
	at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:573)
	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1227)
	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1094)
	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1017)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1899)
	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:3048)



} -Stdout ###

```

xBis7:
I've commented out the two failing tests. After that, all other tests for both Trino and Spark are passing.

PR #25, opened 2024-04-18
gr-oss-661. Separate out our ranger local configs so they don't get passed in the diff file.

xBis7:
There are two changes in this PR
* Applying a Ranger patch that simplifies the docker env (and cleaning it up)
* Deleting the tarballs in the `ranger/dev-support/ranger-docker/dist` dir instead of building the project
  * `ranger_in_docker` will build the project
  * without this change, there is a corner case where we end up building Ranger twice


simplify_ranger_docker.diff
mladjan-gadzic, line 1: Is this needed?

mladjan-gadzic, line 1: Why do we need this in this repo?

xBis7: I removed these changes from the Ranger repo so that we don't pass them to the BigData team. We apply this patch right before building Ranger, every time.

> The big data team won't be expecting these changes and may be confused about them. I would prefer it if we make our ranger specific config changes with a patch file in our build, something like this:
> 
> https://github.com/xBis7/docker-setup-helper-scripts/blob/8b2b3a42a0612e9d1e66bbb672dcf6dce4c3628c/trino-hms-hdfs-ranger/setup/build_projects.sh#L73
> 
> But this would not be a user defined patch file, it would be a part of the repo and would add our specific ranger configs.

from https://github.com/G-Research/gr-oss/issues/661

xBis7:
I was able to run `run_all.sh` successfully in a clean directory. The latest revision should look good.

GeorgeJahad:
This PR requires a new version of ranger-2.4-with-hmsa that breaks the old versions of the docker script.  I think it would be better to use a new branch, something like ranger-2.4-with-hmsa-v2, instead of breaking the old scripts.

GeorgeJahad:
I would also revert the changes to ranger-2.4-with-hmsa so that the old scripts start working again.

xBis7:
@GeorgeJahad Which version of the scripts breaks? Are you referring to the docker changes that we are patching?

xBis7:
@GeorgeJahad I've made the changes. I started in a clean directory and ran `test_all.sh` without any issues.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -11,8 +11,8 @@ PROJECT_SPARK="spark"
 CURRENT_REPO="docker-setup-helper-scripts"
 
 # Project branches
-RANGER_BRANCH="ranger-2.4-with-hmsa"
-HIVE_BRANCH=
+RANGER_COMMIT_SHA="2f7682e412caf81202e223bd2fe9c50b5c1188f4"
```
GeorgeJahad, line 14: This is the hive sha isn't it?

xBis7: I copy pasted it from the Hive3 configuration. I must have copied the wrong value. I'll fix it.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -11,8 +11,8 @@ PROJECT_SPARK="spark"
...
+HIVE_COMMIT_SHA=
```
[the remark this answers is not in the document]
xBis7: Yes, whether we have Hive3 or we have Hive4, this is going to get a value.

We should probably leave `RANGER_COMMIT_SHA=` from above, empty as well.

PR #26, opened 2024-04-22
gr-oss-643

mladjan-gadzic:
This test was commented out. In practice it only affects Hive 4. It is better to have it uncommented and executed for Hive 3 and fail for Hive 4 than not have it in the test suite for Hive 3.

### DIFF
--- a/trino-hms-hdfs-ranger/tests/trino/5_test_trino_select_alter_perm.sh
+++ b/trino-hms-hdfs-ranger/tests/trino/5_test_trino_select_alter_perm.sh
@@ -49,12 +49,12 @@ retryOperationIfNeeded "$abs_path" "alterTrinoTable $TRINO_TABLE $NEW_TRINO_TABL
 # successMsg="INSERT: 1 row"
 # retryOperationIfNeeded "$abs_path" "performTrinoCmd $cmd" "$successMsg" "false"
 
-# TODO: Fix this test
+# FIXME: Comment out this test for Hive 4
 # It fails with the following error: "Query 20240402_105548_00016_g47rj failed: Cannot delete from non-managed Hive table".
 # It should not fail because $TABLE_ANIMALS is created in Hive managed space.
-#echo ""
-#echo "- INFO: Delete from $TABLE_ANIMALS table."
-#echo "- INFO: [alter] should succeed."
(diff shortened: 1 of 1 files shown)


trino-hms-hdfs-ranger/tests/trino/5_test_trino_select_alter_perm.sh
```diff
@@ -49,12 +49,12 @@ retryOperationIfNeeded "$abs_path" "alterTrinoTable $TRINO_TABLE $NEW_TRINO_TABL
...
+# FIXME: Comment out this test for Hive 4
 # It fails with the following error: "Query 20240402_105548_00016_g47rj failed: Cannot delete from non-managed Hive table".
 # It should not fail because $TABLE_ANIMALS is created in Hive managed space.
-#echo ""
-#echo "- INFO: Delete from $TABLE_ANIMALS table."
-#echo "- INFO: [alter] should succeed."
-#cmd="delete from hive.default.$TABLE_ANIMALS;"
-#successMsg="DELETE"
-#retryOperationIfNeeded "$abs_path" "performTrinoCmd $cmd" "$successMsg" "false"
+echo ""
```
xBis7, line 55: Instead of commenting the test for Hive4, we can choose whether to run it or not, depending on the Hive version.

```suggestion
  if [ "$HIVE_VERSION" != "4" ]; then
    echo ""
    echo "- INFO: Delete from $TABLE_ANIMALS table."
    echo "- INFO: [alter] should succeed."
    cmd="delete from hive.default.$TABLE_ANIMALS;"
    successMsg="DELETE"
    retryOperationIfNeeded "$abs_path" "performTrinoCmd $cmd" "$successMsg" "false"
  fi
```

Because we never check for `3` but we check whether `4` exists or not, the check should be `!= 4` and not `== 3`.

`HIVE_VERSION` will be available here, right?

mladjan-gadzic: Handy! It works, I've tested it out.

xBis7 (approved):
@mladjan-gadzic Thanks for the changes!

mladjan-gadzic:
@xBis7 thanks for a quick review!

PR #28, opened 2024-04-29
Update ranger commit SHA

xBis7:
Updated the commit SHA for Ranger. The new Ranger version contains fixes for missing jars for `ranger-admin` and `ranger-usersync`. 

Hadoop 3.3.6 uses a shaded version of guava from hadoop thirdparty. We have to replace guava with the shaded version.

`ranger-usersync` was failing to start due to the guava dependency but because we were using a simplified ranger docker env that didn't have `ranger-usersync`, we didn't see this error. As part of this PR, I also removed the Ranger patch that simplifies the docker env. We should be starting all services to make sure there are no more errors.

I've run all scripts in a clean directory without any errors.


simplify_ranger_docker.diff
```diff
@@ -1,119 +0,0 @@
...
-             fi
-         else
-             cd ${RANGER_HOME}
--            mvn clean package -DskipTests
--            if [ $? -ne 0 ]
--            then
--                echo "ERROR: Unable to complete RANGER build. Exiting ...."
--                exit 1
--            fi
-+#            mvn clean package -DskipTests
```
GeorgeJahad, line 111: So now it looks like we are getting a second full ranger build.  Is it because this is no longer commented out?

xBis7: In the past few days, I have been using the patched code and instead of building the project, I was deleting the tarballs and then running `./ranger_in_docker up`. The script would then build the project for me.

After a bit, I realized that this shouldn't happen because we are supposed to have commented out the building part. Therefore, some other part of the code is responsible for the second build.

I looked more into it and the second build is coming from executing `ranger-build.sh`.

This is a simplified version of the code,

```
if [ ${DOCKER_MAVEN_BUILD} -eq 0 ]
then
  docker-compose -f docker-compose.ranger-base.yml -f docker-compose.ranger-build.yml down --remove-orphans

  docker-compose -f docker-compose.ranger-base.yml -f docker-compose.ranger-build.yml up
else
  mvn clean package -DskipTests
fi
```

We commented out the `else` block which is run if `DOCKER_MAVEN_BUILD` = 1.

Above, there is this check

```
if [ -z "${DOCKER_MAVEN_BUILD}" ]
then
    DOCKER_MAVEN_BUILD=0
fi
```
which means that if `DOCKER_MAVEN_BUILD` is null or empty, set it to 0. We never set the env variable `DOCKER_MAVEN_BUILD` and therefore it's always empty.

In our setup, we are actually always going over the `if` block and not the `else` block. The build is due to this cmd

```
docker-compose -f docker-compose.ranger-base.yml -f docker-compose.ranger-build.yml up
```

`docker-compose.ranger-build.yml` uses `Dockerfile.ranger-build` which calls `ranger-build.sh` which is in fact building the project for the second time.

xBis7:
@GeorgeJahad Let's leave this PR open for a while. Once we have a fully working Ranger SHA, I'll update the commit in the scripts and in the meantime I'll push a new script.

There is a process that I find myself constantly repeating while testing Ranger
* Delete the dist jars
* Delete the Ranger images
* Run `./ranger_in_docker up`

I'll add a script that automates that.

xBis7:
I've run the tests for the latest SHA and they are all passing.

GeorgeJahad (approved):

xBis7:
Thanks @GeorgeJahad for the review. I'm merging it.

PR #31, opened 2024-05-14
Update hive commit SHA

xBis7:
I've replaced the last Hive commit which was squash merged with individual commits and also merged the UGI fix. 

Therefore the current commit is invalid, doesn't exist anymore and needs to be replaced. Also, the new commit contains and another fix.


mladjan-gadzic (approved):

mladjan-gadzic:
@xBis7 thanks for merging UGI workaround as well!

PR #33, opened 2024-05-30
Scripts for managing policies by making calls to the Ranger API.

xBis7:
https://github.com/G-Research/gr-oss/issues/713

This new directory is independent and can be copied under any system running Ranger and be used. In the terminal session use this script first to set your Ranger variables and then use the rest of the scripts.
```
❯ ./ranger_api/export_ranger_ui_variables.sh <username> <password> <hostname> <port>
```

The scripts right now use the default policy names that Ranger sets. It can be later parameterized but there is no need to over complicate the parameters right now.

Here are some example commands for manually testing the scripts while checking the Ranger UI. After each API call, you need to refresh the UI.

```
- Set up the env and load a psql dumb file.
❯ ./setup/setup_for_trino_spark_testing.sh /home/projects

- Check the UI to see that all policies exist.

- Delete all policies.
❯ ./ranger_api/delete_policy.sh hdfs all

❯ ./ranger_api/delete_policy.sh hive all_db

❯ ./ranger_api/delete_policy.sh hive defaultdb

❯ ./ranger_api/delete_policy.sh hive url

- Create them again with new values.
❯ ./ranger_api/create_update/create_update_hdfs_path_policy.sh "/*" "read,write:hadoop" "create"

❯ ./ranger_api/create_update/create_update_hive_all_db_policy.sh "select,drop:hadoop/alter,read:spark" "create"

❯ ./ranger_api/create_update/create_update_hive_defaultdb_policy.sh "select,drop:hadoop/alter,read:spark" "create"

❯ ./ranger_api/create_update/create_update_hive_url_policy.sh "read:spark" "create"

- Update all policies with more or less allow conditions.
❯ ./ranger_api/create_update/create_update_hdfs_path_policy.sh "/dir1,/test/*" "read,write:hadoop/execute:spark" "put"

❯ ./ranger_api/create_update/create_update_hive_all_db_policy.sh "select,alter:trino" "put"

❯ ./ranger_api/create_update/create_update_hive_defaultdb_policy.sh "select,index,lock:spark" "put" "test_col"

❯ ./ranger_api/create_update/create_update_hive_url_policy.sh "read:spark/write:hadoop" "put"

```


mladjan-gadzic:
`./ranger_api/export_ranger_ui_variables.sh <username> <password> <hostname> <port>`

This command does not work for me. Variables are not expored. It seems that those are only exported within session of a shell script, not within terminal session where that script is executed from.

xBis7:
>  I cannot only run this script. Instead I need to run it with source SCRIPT_PATH.

So instead of

`./ranger_api/export_ranger_ui_variables.sh <username> <password> <hostname> <port>`

you have to add the values that you need and run this

`source ranger_api/export_ranger_ui_variables.sh`

is that correct? Does it work that way?

trino-hms-hdfs-ranger/ranger_api/export_ranger_ui_variables.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
```
xBis7, line 2: @mladjan-gadzic What about adding a comment here that this file needs to be modified with the necessary values and then run with `source`?

mladjan-gadzic: I am fine with passing values through parameters and just running with `source`. So no worries, I just wanted to update you with my findings.

xBis7: I haven't used this script. I added a comment just to know.

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+  #     ],
+  #     "isExcludes":false,
+  #     "isRecursive":false
+  #   }
+  # },
+  #
+  # 'jq' will be used by the caller to prettify the result.
+
+  # Set IFS to a comma.
+  IFS=','
```
GeorgeJahad, line 114: Can we get rid of IFS and just use jq?

xBis7: But the IFS has to do with the user input and not the json. It's for splitting whatever the user provides. I will look into it.

xBis7: We can do that by using json files for the policies. Instead of providing script parameters like so `select,drop:hadoop/alter,read:spark` and then splitting them with IFS, we will edit the json files and add the info there. The scripts will read the json and interpret them with `jq` and will update the ranger_api.

For the above example, the json file will look like this
```json
[
  {
    "operations": ["select", "drop"],
    "users": ["hadoop"]
  },
  {
    "operations": ["alter", "read"],
    "users": ["spark"]
  }
]
```

What do you think? Is this better?

GeorgeJahad: no I agree.  let's leave it for now.

GeorgeJahad:
I think the example commands above should be a comment in the code somewhere. instead of a comment in this pr.  people will want to see it after the pr is merged.

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+
+  curl -iv -u "$RANGER_UI_USERNAME":"$RANGER_UI_PASSWORD" -X DELETE http://$RANGER_UI_HOSTNAME:$RANGER_UI_PORT/service/public/v2/api/policy/$id
+}
+
+getIdFromRangerPolicyJsonRes() {
+  res=$1
+
+  # Example: "id":22,
+  # and we need to get just the number.
+  id=$(echo "$res" | grep -o '"id":[0-9]*' | awk -F: '{print $2}')
```
GeorgeJahad, line 61: wouldn't this be better done with jq?

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+  echo "$id"
+}
+
+getGuidFromRangerPolicyJsonRes() {
+  res=$1
+  
+  # Example: "guid":"d0be1b59-31be-4d0f-94e9-e936f25d0794"
+  # We need to get everything between the double quotes.
+  # (tr -d '"') trims it and removes the quotes.
+  guid=$(echo "$res" | grep -o '"guid":"[^"]*"' | awk -F':' '{print $2}' | tr -d '"')
```
GeorgeJahad, line 72: wouldn't this be better done with jq?

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+}
+
+# If the policy has custom resources, then there is a resource signature.
+# We can perform api calls without specifying this.
+# This method is currently not used. I'm leaving it here in case it's needed.
+getResourceSignature() {
+  res=$1
+  
+  # Example: "resourceSignature":"a6a328c623a4eb15d84ccf34df65c83acba3c59c7df4a601780b1caa153aa870"
+  resource_sig=$(echo "$res" | grep -o '"resourceSignature":"[^"]*"' | awk -F':' '{print $2}' | tr -d '"')
```
GeorgeJahad, line 84: wouldn't this be better done with jq?

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+RANGER_UI_PORT=
+
+if [[ "${USE_RANGER_UI_CUSTOM_VALUES}" != "true" ]]; then
+  RANGER_UI_USERNAME="admin"
+  RANGER_UI_PASSWORD="rangerR0cks!"
+  RANGER_UI_HOSTNAME="localhost"
+  RANGER_UI_PORT="6080"
+fi
+
+getRangerPolicyJsonRes() {
```
GeorgeJahad, line 20: Let's update everything to use "Response" instead of "Res".  Res is confusing to me.

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+  res=$1
+  
+  # Example: "resourceSignature":"a6a328c623a4eb15d84ccf34df65c83acba3c59c7df4a601780b1caa153aa870"
+  resource_sig=$(echo "$res" | grep -o '"resourceSignature":"[^"]*"' | awk -F':' '{print $2}' | tr -d '"')
+
+  echo "$resource_sig"
+}
+
+getResourcesJsonArray() {
+  resource_values=$1
```
GeorgeJahad, line 90: what is the format of resources values? it looks like it is supposed to be the "resources example" below?

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+  array+="]"
+
+  # Reset IFS to its default value.
+  IFS=' '
+
+  echo -e "$array"
+}
+
+getAccessesJsonArray() {
+  accesses=$1
```
GeorgeJahad, line 141: document the parameter please

GeorgeJahad: Also what is the difference between this method and the previous one

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,321 @@
...
+  done
+  accesses_array+="]"
+
+  # Reset IFS to its default value.
+  IFS=' '
+
+  echo -e "$accesses_array"
+}
+
+getUsersJsonArray() {
```
GeorgeJahad, line 187: how does this method differ from the others?

trino-hms-hdfs-ranger/ranger_api/create_update/create_update_hdfs_path_policy.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
+source "./ranger_api/lib.sh"
+
+set -e
+
+resources_path=$1
+
+# It can be like this for multiple conditions:
```
GeorgeJahad, line 9: instead of "It" say "policy_items" so we know what "it" is.

trino-hms-hdfs-ranger/ranger_api/create_update/create_update_hdfs_path_policy.sh
```diff
@@ -0,0 +1,83 @@
...
+# accesses1:users1/accesses2:users2/accesses3:users3
+# If there is only 1 allow condition, then it will be just: accesses1:users1
+policy_items=$2
+request_type=$3
+
+policy_name="all - path"
+policy_uri_name="all%20-%20path"
+
+json_payload+=$(cat <<EOF
```
GeorgeJahad, line 18: is this just appending a "{"?  wouldn't it be clearer to just do that?

trino-hms-hdfs-ranger/ranger_api/lib.sh
GeorgeJahad: getPolicyItemsJsonArray does have multiple delimiters, so it makes sense to use IFS there.  For all the methods parsing "," delimiters, let's just use jq.

GeorgeJahad: doesn't this do what we need:
```
  echo "a,b,c" | jq -R  'split(",")'

[
  "a",
  "b",
  "c"
]
```

trino-hms-hdfs-ranger/ranger_api/lib.sh
xBis7: You are right. Done.

trino-hms-hdfs-ranger/ranger_api/lib.sh
xBis7: Done.

trino-hms-hdfs-ranger/ranger_api/lib.sh
xBis7: Done.

trino-hms-hdfs-ranger/ranger_api/create_update/create_update_hdfs_path_policy.sh
xBis7: Sure. Changed it.

trino-hms-hdfs-ranger/ranger_api/lib.sh
xBis7: You are right, they have the same implementation. I'll combine them.

trino-hms-hdfs-ranger/ranger_api/lib.sh
xBis7: Thanks! I've used `echo "a,b,c" | jq -R  'split(",")'` to replace all IFS usages.

xBis7:
I'll push a change that simplifies the output of each command.

xBis7:
I've updated every curl command so that it only returns a status code. The status code is then checked for every call.

Example output
```
❯ ./ranger_api/create_update/create_update_hdfs_path_policy.sh "/dir1,/test/*" "read,write:hadoop/execute:spark" "put"

-- Updating Ranger policy: hdfs / all - path

Status code: OK
```

OK means that the returned code matches the regex `2[0-9][0-9]`.

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
```diff
@@ -0,0 +1,40 @@
...
+
+prepare_env=$1
+abs_path=$2
+
+# Set up the env and load a psql dumb file.
+if [ "$prepare_env" == "true" ]; then
+  ./setup/setup_for_trino_spark_testing.sh "$abs_path"
+fi
+
+# Delete all policies.
```
GeorgeJahad, line 15: can we read the deleted policies back and confirm they are what we expect?

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
```diff
@@ -0,0 +1,40 @@
...
+./ranger_api/delete_policy.sh "hdfs" "all"
+
+./ranger_api/delete_policy.sh "hive" "all_db"
+
+./ranger_api/delete_policy.sh "hive" "defaultdb"
+
+./ranger_api/delete_policy.sh "hive" "url"
+
+# Create the policies again with new values.
```
GeorgeJahad, line 24: can we read the changes back and confirm they are what we expect?

trino-hms-hdfs-ranger/ranger_api/create_update/create_update_hdfs_path_policy.sh
```diff
@@ -0,0 +1,86 @@
...
+# If there is only 1 allow condition, then it will be just: accesses1:users1
+policy_items=$2
+request_type=$3
+
+policy_name="all - path"
+policy_uri_name="all%20-%20path"
+
+json_payload+="{"
+
+# If it's a create request, then we need to include the id and the guid in the json.
```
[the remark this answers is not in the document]
xBis7: You are right. Fixed.

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
GeorgeJahad: https://stackoverflow.com/questions/31930041/using-jq-or-alternative-command-line-tools-to-compare-json-files

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
GeorgeJahad: sure

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
xBis7: ok, I'll look into it.

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
xBis7: ok, I'll look into it.

xBis7: Done.

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
xBis7: As far as I can see, using jq to compare the json results means that we would have to write 8 json files and use them for comparison.

xBis7: I think I found a workaround. We can get the fields one by one.

xBis7: For these parameters
`"/dir1,/test/*" "read,write:hadoop/execute:spark" "put"`

The check will be like this

```shell
response=$(getRangerPolicyJsonResponseUsingShortNames "hdfs" "all")

echo "$response" | jq '.resources.path | if .values[0] == "/dir1" then "Success" else "Failure" end'
echo "$response" | jq '.resources.path | if .values[1] == "/test/*" then "Success" else "Failure" end'

echo "$response" | jq '.policyItems[0].accesses[0] | if .type == "read" then "Success" else "Failure" end'
echo "$response" | jq '.policyItems[0].accesses[1] | if .type == "write" then "Success" else "Failure" end'

echo "$response" | jq '.policyItems[0] | if .users[0] == "hadoop" then "Success" else "Failure" end'

echo "$response" | jq '.policyItems[1].accesses[0] | if .type == "execute" then "Success" else "Failure" end'

echo "$response" | jq '.policyItems[1] | if .users[0] == "spark" then "Success" else "Failure" end'
```

GeorgeJahad: > As far as I can see, using jq to compare the json results means that we would have to write 8 json files and use them for comparison.

is that bad?

xBis7: > > As far as I can see, using jq to compare the json results means that we would have to write 8 json files and use them for comparison.
>
> is that bad?

It's too much and it's not expandable. But if I create some parameterized methods with the approach in this comment 

> For these parameters
> "/dir1,/test/*" "read,write:hadoop/execute:spark" "put"
> 
> The check will be like this
> 
> response=$(getRangerPolicyJsonResponseUsingShortNames "hdfs" "all")
> 
> echo "$response" | jq '.resources.path | if .values[0] == "/dir1" then "Success" else "Failure" end'
> echo "$response" | jq '.resources.path | if .values[1] == "/test/*" then "Success" else "Failure" end'
> 
> echo "$response" | jq '.policyItems[0].accesses[0] | if .type == "read" then "Success" else "Failure" end'
> echo "$response" | jq '.policyItems[0].accesses[1] | if .type == "write" then "Success" else "Failure" end'
> 
> echo "$response" | jq '.policyItems[0] | if .users[0] == "hadoop" then "Success" else "Failure" end'
> 
> echo "$response" | jq '.policyItems[1].accesses[0] | if .type == "execute" then "Success" else "Failure" end'
> 
> echo "$response" | jq '.policyItems[1] | if .users[0] == "spark" then "Success" else "Failure" end'

we should be good and the methods will be reusable. We won't have to create new set of json files for every test.

GeorgeJahad: ok

mladjan-gadzic: @xBis7 this is still to be implemented, right?

mladjan-gadzic:
@xBis7 let me know once final commits are in so I can test this against c3 env.

xBis7:
> @xBis7 let me know once final commits are in so I can test this against c3 env.

Any new commits will be related to the test and checking the Ranger Admin data after running each script. The implementation for the scripts won't change.

trino-hms-hdfs-ranger/ranger_api/test_ranger_api_calls.sh
xBis7: I've pushed the changes. This is an example output of the test

```
-- Updating Ranger policy: hdfs / all - path

Status code: OK

Success - resource_path: /dir1
Success - resource_path: /test/*
Success - policy_access: read
Success - policy_access: write
Success - policy_user: hadoop
Success - policy_access: execute
Success - policy_user: spark

-- Updating Ranger policy: hive / all - database, table, column

Status code: OK

Success - policy_access: select
Success - policy_access: alter
Success - policy_user: trino

-- Updating Ranger policy: hive / default database tables columns

Status code: OK

Success - resource_column: test_col
Success - policy_access: select
Success - policy_access: index
Success - policy_access: lock
Success - policy_user: spark
```

xBis7:
@mladjan-gadzic There are no further changes to be made. It should be good for testing.

trino-hms-hdfs-ranger/ranger_api/lib.sh
```diff
@@ -0,0 +1,410 @@
...
+    res=$(echo "$res_array" | jq ".[$i]" | tr -d '"')
+
+    out=$(echo "$json_response" | jq -r ".resources.$resource_type | if .values[$i] == \"$res\" then \"Success\" else \"Failure\" end")
+    validateResult "$out" "resource_$resource_type: $res"
+  done
+}
+
+iterateAndValidatePolicies() {
+  json_response=$1
+  policies=$2
```
GeorgeJahad, line 351: Please document the format of this parameter.  The rest looks good.  I'll approve so that you can merge once you add the document.

GeorgeJahad (approved):

trino-hms-hdfs-ranger/ranger_api/lib.sh
xBis7: Done.

PR #38, opened 2024-06-21
Enable Trino-Hive impersonation and replace psql dump files with ranger-api calls

xBis7:
* enabled trino-hive impersonation so that we can use multiple users with trino
* replaced all psql dump file loads with ranger-api calls
* deleted all psql dump files except two that are used as a base

I've run all tests for trino and spark at least 5 times in a row. No issues.


trino-hms-hdfs-ranger/tests/spark/database/1_create_database_no_perms.sh
```diff
@@ -12,6 +12,10 @@ echo "- INFO: all you need is HDFS perms. No Hive perms are needed."
 echo "- INFO: User [spark] doesn't have HDFS or Hive permissions."
 echo "- INFO: Create database."
 echo "- INFO: User [spark] shouldn't be able to create database."
+
+updateHiveDefaultDbPolicy "select,read:spark,trino"
```
GeorgeJahad, line 16: why aren't there four settings here something like this:
```
+updateHdfsPathPolicy "read,write,execute:hadoop,trino,spark" "/*"
+updateHiveDbAllPolicy "select,update,Create,Drop,Alter,Index,Lock,All,Read,Write,ReplAdmin,Refresh:hive"
+updateHiveDefaultDbPolicy ""
+updateHiveUrlPolicy "select,update,Create,Drop,Alter,Index,Lock,All,Read,Write,ReplAdmin,Refresh:hive"
```

i'm guessing we are either depending on the previous settings, or the other settings don't matter, but in either case I would prefer we make explicit calls everywhere.

GeorgeJahad:
I don't like this ambiguous test names:
```
 echo "### TEST_0 ###"
```
Let's give each test an unambiguous name that is not duplicated anywhere. e.g: 
```
 echo "### TEST_trino_no_hdfs_perms ###"
```

trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
```diff
@@ -10,9 +10,16 @@ abs_path=$1
 
 echo ""
 echo "- INFO: Updating Ranger policies. Users [spark/postgres] will now have all access to Hive default DB."
-./setup/load_ranger_policies.sh "$abs_path" "$HDFS_AND_HIVE_ALL"
```
GeorgeJahad, line 13: when i load the $HDFS_AND_HIVE_ALL into ranger and look at the settings it seem significantly different from the settings below.  Should it be the same for the four policies we are setting.  More details below.

trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
```diff
@@ -10,9 +10,16 @@ abs_path=$1
...
+
+./setup/load_ranger_policies.sh "$abs_path" "$HIVE_BASE_POLICIES"
 waitForPoliciesUpdate
 
+updateHdfsPathPolicy "read,write,execute:hadoop" "/*"
```
GeorgeJahad, line 17: In $HDFS_AND_HIVE_ALL the policy includes postgres and spark

trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
```diff
@@ -10,9 +10,16 @@ abs_path=$1
...
+updateHiveDbAllPolicy "select,update,Create,Drop,Alter,Index,Lock,All,Read,Write,ReplAdmin,Refresh:hive,spark"
```
GeorgeJahad, line 18: In $HDFS_AND_HIVE_ALL the policy includes just hive

trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
```diff
@@ -10,9 +10,16 @@ abs_path=$1
...
+updateHiveDefaultDbPolicy "select,update,Create,Drop,Alter,Index,Lock,All,Read,Write,ReplAdmin,Refresh:trino/select:spark"
```
GeorgeJahad, line 19: In $HDFS_AND_HIVE_ALL the policy sets spark user to have all permissions

trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
```diff
@@ -10,9 +10,16 @@ abs_path=$1
...
+updateHiveUrlPolicy "select,update,Create,Drop,Alter,Index,Lock,All,Read,Write,ReplAdmin,Refresh:hive,trino,spark"
```
GeorgeJahad, line 20: In $HDFS_AND_HIVE_ALL the policy just sets the hive user.

xBis7:
> I don't like this ambiguous test names:
> 
>  echo "### TEST_0 ###"
> Let's give each test an unambiguous name that is not duplicated anywhere. e.g:
> 
>  echo "### TEST_trino_no_hdfs_perms ###"

The naming was like this so that we could easily match the test number with the psql dump file number. Now that all psql dump files have been removed, I'll change the messages to the test names.

trino-hms-hdfs-ranger/tests/spark/database/1_create_database_no_perms.sh
xBis7: We are reusing all the other policies. I'll update it accordingly.

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/4_test_alter_table_uri.sh
```diff
@@ -40,7 +40,7 @@ retryOperationIfNeeded "$abs_path" "runSparkTest $SPARK_TEST_FOR_EXCEPTION_FILEN
 echo ""
 echo "Creating Hive URL policies again."
 
-./ranger_api/create_update/create_update_hive_url_policy.sh "read,write:spark" "create"
+createHiveUrlPolicy "read,write:spark"
```
xBis7, line 43: @GeorgeJahad I have two questions.

1. There are tests like this one where we are reusing policies from a previous test and therefore we aren't loading any policies at all. Do you want me to add the previous policies on top so that it's clear which policies it's using?
2. This test is deleting the Hive URL policies and then creating them again. Do you want me to specify the rest of the policies here are well? 

Instead of deleting and creating again, we could be updating them twice. The first time to empty values and the second to the create values.

xBis7: > Instead of deleting and creating again, we could be updating them twice. The first time to empty values and the second to the create values.

I think this is cleaner. I'm going to make this change and we can later revert it.

trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
xBis7: I updated it to match the policies from the psql dump. TBH, I didn't pay much attention to these policies because `setup_for_testing_hms_policies.sh` isn't used by any other file. I use that and `setup_for_testing_hms_url_policies.sh` as a starting point for manual testing.

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
```diff
@@ -7,9 +7,14 @@ set -e
 abs_path=$1
 
 echo ""
-echo "- INFO: Updating Ranger policies. User [postgres] will now have only [select] access to Hive default DB."
+echo "- INFO: Updating Ranger policies. User [trino] will now have only [select] access to Hive default DB."
 echo "- INFO: HDFS access for user [trino] has been removed."
-./setup/load_ranger_policies.sh "$abs_path" "$HDFS_AND_HIVE_SELECT"
+
+updateHdfsPathPolicy "read,write,execute:hadoop,spark" "/*"
```
xBis7, line 13: User `postgres` exists here in the psql dump file. User `postgres` has been replaced by user `trino`. Later on, in this test we are trying to run `insert into` which is a pure HDFS action that used to be performed by user `trino` because hive wasn't involved. If we replace user `postgres` with user `trino`, the `insert into` won't fail with an HDFS error as expected but instead it will work.

GeorgeJahad:
Please fix these "echo"'s to include the component name:
```
find . -type f | xargs grep Test[0-9]
./trino-hms-hdfs-ranger/tests/spark/hive_url_policies/1_test_create_database.sh:echo "Test1: ############### create database without and with Hive URL policies ###############"
./trino-hms-hdfs-ranger/tests/spark/hive_url_policies/2_test_select_drop_with_a_second_user.sh:echo "Test2: ############### test select and drop with user 'games' ###############"
./trino-hms-hdfs-ranger/tests/spark/hive_url_policies/3_test_create_table_managed_and_not_managed.sh:echo "Test3: ############### create table (managed + not-managed) without and with Hive URL policies ###############"
./trino-hms-hdfs-ranger/tests/spark/hive_url_policies/4_test_alter_table_uri.sh:echo "Test4: ############### rename table location without and with Hive URL policies ###############"
./trino-hms-hdfs-ranger/tests/trino/hive_url_policies/2_test_select_drop_with_a_second_user.sh:echo "Test2: ############### test select and drop with user 'games' ###############"
./trino-hms-hdfs-ranger/tests/trino/hive_url_policies/1_test_create_schema.sh:echo "Test1: ############### create schema without and with Hive URL policies ###############"
./trino-hms-hdfs-ranger/tests/trino/hive_url_policies/3_test_create_table_managed_and_not_managed.sh:echo "Test3: ############### create table (managed + not-managed) without and with Hive URL policies ###############"
./trino-hms-hdfs-ranger/tests/trino/hive_url_policies/4_test_alter_table_uri.sh:echo "Test4: ############### rename table location without and with Hive URL policies ###############"
```

GeorgeJahad (approved):
LGTM.  Please address this final comment and then merge: https://github.com/xBis7/docker-setup-helper-scripts/pull/38#issuecomment-2190206935

PR #39, opened 2024-06-26
Simple load tests

xBis7:


trino-hms-hdfs-ranger/load_testing/env_variables.sh
```diff
@@ -0,0 +1,49 @@
...
+HIVE_USER="hive"
+SPARK_USER1="spark"
+SPARK_USER2="test1"
+SPARK_USER3="test2"
+SPARK_USER4="test3"
+
+# Hostnames.
+NAMENODE_HOSTNAME="docker-namenode-1"
+DN1_HOSTNAME="docker-datanode-1"
+DN2_HOSTNAME="docker-datanode-2"
```
GeorgeJahad, line 19: We don't have access to some of these hosts in c3, (like dn2 and dn3.)  It doesn't seem like they get used anywhere in this code, are they?

I would say any hostnames that are not used should be removed.  That way it is clear what is required for the tests in c3.

trino-hms-hdfs-ranger/load_testing/run_test.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+if [ "$run_test_file3" == 0 ]; then
+
+  runInsertSelectTableOnRepeatWithAccess "$SPARK_USER3" "$iteration_num" "$background_run"
+fi
+
+if [ "$run_test_file4" == 0 ]; then
+
+  runInsertSelectTableOnRepeatNoAccess "$SPARK_USER4" "$iteration_num" "$background_run"
```
GeorgeJahad, line 60: This would be better as a bash case statement wouldn't it?

xBis7: aah yes, you are right. It started like this because I was thinking about running multiple tests at once. That's why I added a flag for each test.

trino-hms-hdfs-ranger/load_testing/env_variables.sh
xBis7: I deleted all hostnames that aren't used.

trino-hms-hdfs-ranger/load_testing/run_test.sh
xBis7: I cleaned it up.

GeorgeJahad (approved):
lgtm, let's merge it and rebase the other pr.

xBis7:
Thanks @GeorgeJahad !

PR #44, opened 2024-07-04
Load test for running parallel Spark jobs on large datasets

xBis7:
I've pushed a set of files with 10k lines. I have two more sets, one with 50k lines and one with 500k lines. I haven't pushed the larger sets because of their size but I can share a tarball with them.

10k and 50k finish pretty fast on docker. 500k takes 1 or 2 minutes. This probably won't be the case with c3.

I don't know what else needs to be parameterized for c3.


mladjan-gadzic:
I though that running a Spark job is different than passing Scala file to Spark shell. Am I wrong?

xBis7:
> I though that running a Spark job is different than passing Scala file to Spark shell. Am I wrong?

As far as I know, you have 3 ways to run something on spark
* run commands directly on the `spark-shell`
* submit a scala file with commands on the `spark-shell`
* write a Java, Scala or Python program that makes calls to the corresponding Spark APIs and executes commands.

A spark job is any computation that spark performs. Instead of performing DDL operations, we can load the entire table in a data frame and then do complex queries, use the results to create new tables and then write the data back to HDFS. That's a spark job. It doesn't matter how you run it in spark. It can be through a scala file or by executing the commands one by one directly in the shell.

https://sparkbyexamples.com/spark/what-is-spark-job/

trino-hms-hdfs-ranger/big-data-c3-tests/load-testing/run_spark_jobs.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+source "./big-data-c3-tests/env_variables.sh"
+source "./big-data-c3-tests/lib.sh"
+
+set -e
+
+abs_path=$1
+setup=$2
+dataset_size=$3
```
GeorgeJahad, line 10: comments.  we need to do a better job on parameters.

trino-hms-hdfs-ranger/big-data-c3-tests/load-testing/scala-test-files/spark_job_on_large_dataset.scala
```diff
@@ -0,0 +1,172 @@
+
+object Utils {
+
```
GeorgeJahad, line 3: How about a paragraph comment here on what this test does.

trino-hms-hdfs-ranger/big-data-c3-tests/load-testing/run_spark_jobs.sh
xBis7: I've added comments.

trino-hms-hdfs-ranger/big-data-c3-tests/load-testing/scala-test-files/spark_job_on_large_dataset.scala
xBis7: I've added the paragraph.

xBis7:
@GeorgeJahad I've addressed your comments, rebased the branch and resolved conflicts with master.

trino-hms-hdfs-ranger/big-data-c3-tests/load-testing/run_spark_jobs.sh
```diff
@@ -0,0 +1,77 @@
...
+# param 'abs_path':
+#   projects path, used for setup and copying files under the Spark container
+abs_path=$1
+# param 'setup':
+#   If true, the script will start with a fresh environment,
+#   else, the script will delete the data generated by a previous run.
+#   This parameter helps run the script multiple times in a row.
+setup=$2
+# param 'dataset_size':
+#   Under directory compose/hadoop/docker/load_test_data, there are directories with
```
GeorgeJahad, line 17: 50 and 500 are not in that directory, (and i don't want to add them.)  Are they in a tarball somewhere?  Can you update this comment to clarify where they are?

GeorgeJahad (approved):
lgtm.  please update the one comment I mentioned before merging.

trino-hms-hdfs-ranger/big-data-c3-tests/load-testing/run_spark_jobs.sh
xBis7: I updated the comment to provide the info.

xBis7:
@GeorgeJahad I updated the comment. I will merge the PR. Thanks for the review!

GeorgeJahad:
Thanks @xBis7 

Where is the tarball?

GeorgeJahad:
The tarball is here:

https://files.slack.com/files-pri/TGH6PQ1ED-F07CP9A671C/download/load_test_data.tar.xz?origin_team=TGH6PQ1ED

PR #45, opened 2024-07-05
Initialize the 'externalCatalog' object in advance, when needed to avoid defaultdb SELECT error

xBis7:
https://github.com/G-Research/gr-oss/issues/758

The first time that we are running a command on an external db in a `spark-shell` instance, the `externalCatalog` object has to be initialized.

The initialization checks if the default db exists and if it needs to be created. That requires SELECT access on the default db and for that reason we might be getting a defaultdb SELECT error on an external db operation.

The initialization needs to happen once for every new `spark-shell` instance.

To workaround this, we need to
* provide select access 
* start a new `spark-shell` instance
* run any command on the external catalog
* remove the select access
* run the test command on the same `spark-shell`

We need a shell script running in the background and making the select updates while the test scala file runs on the `spark-shell` and executes the commands.

To achieve synchronization between the two and let them know when to execute what, I've used tmp files to signal the two processes.

The flow goes like this
* Start the shell method in the background
* Submit the scala file in the `spark-shell`
  * This is blocked and waits for a signal file that the initial select update has been made
* The scala file is blocked and waiting, the shell method starts first
* The shell method 
  * updates the defaultdb policies and provides select access
  * creates the signal file to let the scala file know to start the execution
  * waits until a signal file is created by the scala file
* The scala file
  * waits for the predefined interval to make sure that the policies are updated
  * runs `show databases` to initialize the object
  * creates the signal file to let the shell method know to remove the select access
  * waits until the shell method has created a signal file that the update has been made
* The shell method
  * picks up that there is a signal file to remove the select access
  * removes the select access
  * creates the signal file that the update has been made
  * FINISHES
* The scala file
  * picks up that there is a signal file that the update has been made
  * waits for the predefined interval to make sure that the policies are updated
  * FINISHES

The scala file continues with the rest of the test execution, same as before.

### DIFF
--- a/trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
+++ b/trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
@@ -4,6 +4,9 @@ source "./big-data-c3-tests/env_variables.sh"
 
 set -e
 
+UPDATE_SELECT_SIGNAL_FILE="update_select.txt"
+UPDATE_DONE_SIGNAL_FILE="update_done.txt"
+
 copyTestFilesUnderSpark() {
   abs_path=$1
   load_testing=$2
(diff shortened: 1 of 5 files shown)


trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -43,6 +47,59 @@ copyTestFilesUnderSpark() {
   done
 }
 
+# -- RANGER POLICIES --
```
xBis7, line 50: I moved these methods up here, so that they will be available for the new methods.

GeorgeJahad:
We are not going to move forward on this yet.  Let see what the big data team says.  Closing for now.

PR #47, opened 2024-07-08
Cleanup Trino tests

xBis7:
## Changes in this PR

* Replaced all references to the `ACL` term with `POSIX permissions`
* Prefixed all comments referring to the BigData notes with `BigData note:` so that they are easier to spot
* `test_8.sh` creates a `tmp` dir and provides world access to it so that we can remove `/*` from the HDFS policies in this test and all the next ones
* Tried to make tests more consistent with the BigData notes
  * `test_7.sh` changing permissions was redundant
  * `test_9.sh` was updated to get the select error that appears in the notes
  * `test_12.sh - insert into` no longer gets an error on the tmp, it actually gets the EXECUTE error but trino swallows it and throws a TrinoException
  * `test_12.sh - create table` changed policies to be the same as in the notes and changed the expected error
* Removed unecessary changes
* Improved some comments

### DIFF
--- a/trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
+++ b/trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
@@ -10,11 +10,9 @@ echo "## Test 11 ##"
 echo "Attempt various DDL operations against the managed table as a different user"
 echo ""
 
-# BigData note: There are no notes about making policy changes. But if we leave them as they are in the previous test, then user2 will have all HDFS access.
-# We are expecting that 'create table' will fail with an HDFS error which won't happen. Instead, it will fail with a metadata error.
-# "Permission denied: user [$TRINO_USER2] does not have [CREATE] privilege on [gross_test/test2]"
-
-# Remove user2.
+# BigData note: There are no notes about making policy changes.
+# That prevents the create table call below from getting the Execute error it is supposed to get.
+# So we remove HDFS access for user2 here:
 updateHdfsPathPolicy "/data/projects/gross_test,/$HIVE_WAREHOUSE_DIR/gross_test.db" "read,write,execute:$TRINO_USER1"
 
 # BigData note: In order to get the expected errors then user2 must have select access.
--- a/trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
+++ b/trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
@@ -10,14 +10,14 @@ echo "## Test 11 ##"
 echo "Attempt various DDL operations against the managed table as a different user"
 echo ""
 
-# There are no notes about making policy changes. But if we leave them as they are in the previous test, then user2 will have all HDFS access.
+# BigData note: There are no notes about making policy changes. But if we leave them as they are in the previous test, then user2 will have all HDFS access.
 # We are expecting that 'create table' will fail with an HDFS error which won't happen. Instead, it will fail with a metadata error.
 # "Permission denied: user [$TRINO_USER2] does not have [CREATE] privilege on [gross_test/test2]"
 
 # Remove user2.
 updateHdfsPathPolicy "/data/projects/gross_test,/$HIVE_WAREHOUSE_DIR/gross_test.db" "read,write,execute:$TRINO_USER1"
 
-# In order to get the expected errors then user2 must have select access.
+# BigData note: In order to get the expected errors then user2 must have select access.
 # The select that has been added here, isn't part of the BigData notes.
 updateHiveDbAllPolicy "gross_test" "alter,create,drop,index,lock,select,update:$TRINO_USER1/select:$TRINO_USER2"
 
@@ -31,7 +31,7 @@ waitForPoliciesUpdate
 
 # Run the commands as user2.
 command="drop table $TRINO_HIVE_SCHEMA.gross_test.test"
-# In the BigData notes, this command is expected to fail with this error. But this is a Spark error.
+# BigData note: In the notes, this command is expected to fail with this error. But this is a Spark error.
 # The Trino error for lack of DROP privileges, is different.
 
 # expectedMsg="Permission denied: user [$TRINO_USER2] does not have [DROP] privilege on [gross_test/test]"
--- a/trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
+++ b/trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
@@ -10,8 +10,12 @@ echo "## Test 11 ##"
 echo "Attempt various DDL operations against the managed table as a different user"
 echo ""
 
-# It's the same as in the previous test.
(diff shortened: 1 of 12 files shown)


GeorgeJahad:
please list the changes as a comment on the pr to make reviewing easier

xBis7:
> please list the changes as a comment on the pr to make reviewing easier

@GeorgeJahad I updated the PR description.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
```diff
@@ -10,10 +10,14 @@ echo "## Test 11 ##"
 echo "Attempt various DDL operations against the managed table as a different user"
 echo ""
 
-# It's the same as in the previous test.
-updateHdfsPathPolicy "/*,/data/projects/gross_test,/$HIVE_WAREHOUSE_DIR/gross_test.db" "read,write,execute:$TRINO_USER1,$TRINO_USER2"
+# BigData note: There are no notes about making policy changes. But if we leave them as they are in the previous test, then user2 will have all HDFS access.
+# We are expecting that 'create table' will fail with an HDFS error which won't happen. Instead, it will fail with a metadata error.
+# "Permission denied: user [$TRINO_USER2] does not have [CREATE] privilege on [gross_test/test2]"
+
+# Remove user2.
```
GeorgeJahad, line 17: This comment is confusing to me:

```
# BigData note: There are no notes about making policy changes.  But if we leave them as they are in the previous test, then user2 will have all HDFS access.
# We are expecting that 'create table' will fail with an HDFS error which won't happen. Instead, it will fail with a metadata error.
# "Permission denied: user [$TRINO_USER2] does not have [CREATE] privilege on [gross_test/test2]"
# Remove user2.
```

What about something like this instead:
```
# BigData note: There are no notes about making policy changes. That prevents the create table call below from getting the Execute error it is supposed to get.  So we remove HDFS access from user2 here:
```

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_12.sh
```diff
@@ -49,35 +49,41 @@ expectedMsg="test2"
...
+# BigData note: In the notes we are getting this error.
 # expectedMsg="Permission denied: user=$TRINO_USER2, access=EXECUTE, inode=\"/data/projects/gross_test\":"
-expectedMsg="Create temporary directory for hdfs://$NAMENODE_NAME/data/projects/gross_test/test2 failed: Permission denied: user=$TRINO_USER2, access=WRITE, inode=\"/tmp\":"
+expectedMsg="Error moving data files from hdfs://$NAMENODE_NAME/tmp/"
 
-runTrino "$TRINO_USER2" "$command" "shouldPass" "$expectedMsg"
+# BigData note: We can see the EXECUTE error on the namenode logs but Trino swallows the original exception and throws a new one with a new message.
+#
+# 2024-07-09 12:11:56 INFO  Server:3102 - IPC Server handler 8 on default port 8020, call Call#128 Retry#0
+# org.apache.hadoop.hdfs.protocol.ClientProtocol.rename from trino-coordinator-1.rangernw:34720 / 172.19.0.14:34720:
```
GeorgeJahad, line 80: I think it would be clearer if you remove the two lines above, and explicitly say:
```
# This is the error from the namenode log"
```

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_12.sh
```diff
@@ -90,16 +96,17 @@ expectedMsg="Permission denied: user [$TRINO_USER2] does not have [ALTER] privil
 
 runTrino "$TRINO_USER2" "$command" "shouldFail" "$expectedMsg"
 
-# The BigData notes repeat the previous error
+# BigData note: The notes repeat the previous error
```
GeorgeJahad, line 99: Is this comment for the create table op below?  If so, put it by the expected msg down there.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_12.sh
```diff
@@ -90,16 +96,17 @@ expectedMsg="Permission denied: user [$TRINO_USER2] does not have [ALTER] privil
...
 # expectedMsg="Permission denied: user [$TRINO_USER2] does not have [ALTER] privilege on [gross_test/test2]"
-# Assumming that this is a table creation and we expect a metadata error, then it will be a CREATE error.
-# user2 also requires HDFS "write" access so we are adding that here.
-# There won't be metadata access.
-updateHdfsPathPolicy "/*,/data/projects,/data/projects/gross_test,/$HIVE_WAREHOUSE_DIR/gross_test.db" "read,write,execute:$TRINO_USER1/read,write,execute:$TRINO_USER2"
+#
+# This is a 'create table' command where user2 has to create a new directory under '/data/projects'.
+# Based on the provided policies, the user doesn't have write access on '/data/projects'.
+# There should be an HDFS write error.
+updateHdfsPathPolicy "/data/projects/gross_test,/$HIVE_WAREHOUSE_DIR/gross_test.db" "read,write,execute:$TRINO_USER1/read,write,execute:$TRINO_USER2"
```
GeorgeJahad, line 105: why is this policy being added here?  If it is removed, does anything change?

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_8.sh
```diff
@@ -10,15 +10,12 @@ echo "## Test 8 ##"
...
-#
-# ==========================
-# Running trino command: 'insert into hive.gross_test.test values (1, 'Austin')'
-# --------------------------
-#
-# Query 20240704_105000_00008_b6t3t failed: Create temporary directory for hdfs://namenode/opt/hive/data/gross_test.db/test failed: Permission denied: user=trino, access=WRITE, inode="/":hadoop:supergroup:drwxr-xr-x
-updateHdfsPathPolicy "/*,/data/projects/gross_test,/$HIVE_WAREHOUSE_DIR/gross_test.db" "read,write,execute:$TRINO_USER1"
+# BigData note: Create the tmp directory and provide world access to it so that Trino can use it.
+createHdfsDir "tmp"
+changeHdfsDirPermissions "tmp" 777
```
GeorgeJahad, line 15: I'd prefer general setup to go into the setup script, rather adding it to individual tests.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_11.sh
xBis7: I updated it.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_12.sh
xBis7: Done.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_12.sh
xBis7: This policy had a reason of existence before the `tmp` dir change. After providing world-access to the dir, this policy should have been removed. Thanks for catching that.

xBis7:
@GeorgeJahad I've addressed your comments and also rerun all the tests without issues.

PR #48, opened 2024-07-09
Spark and Trino setup for attaching a debugger

xBis7:
For the spark debugger setup to go into effect, you have to rebuild the spark image. The easiest way is to remove the old image and then let docker-compose build it again during the first environment startup.

```
docker image ls | grep spark
```
get the image id

```
docker image rm <id>
```


xBis7:
For the spark debugger setup to work, you have to rebuild the spark image. The easiest way is to remove the old image and then let docker-compose build it again during the first environment startup.

```
docker image ls | grep spark
```
get the image id

```
docker image rm <id>
```

GeorgeJahad:
in general I prefer instructions like the above to be in a readme in the pr, instead of a comment on the pr; (future users will see the readme but won't see the pr.)

mladjan-gadzic:
> in general I prefer instructions like the above to be in a readme in the pr, instead of a comment on the pr; (future users will see the readme but won't see the pr.)

Could you please clarify what did you mean by this? Is it PR description or README file in a repo?

compose/spark/conf/spark-class
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bash
```
mladjan-gadzic, line 1: Why do we need this?

xBis7: I added a comment with an explaination. Check line 20.

GeorgeJahad:
> Could you please clarify what did you mean by this? Is it PR description or README file in a repo?

User documention should be readme's, not comments in the PR description.

GeorgeJahad (dismissed):

GeorgeJahad:
where are the readme changes?

GeorgeJahad:
go ahead and merge after adding the readme

PR #49, opened 2024-07-10
Debug Hadoop

mladjan-gadzic:


mladjan-gadzic:
I've tested it. I am able to connect to the Hadoop NN with a debugger.

xBis7 (approved):
I've tested it locally. LGTM!

PR #54, opened 2024-07-16
Script for debugging a possible drop DB error

xBis7:
context: https://github.com/G-Research/gr-oss/issues/720#issuecomment-2230898954

The script is using methods for spark tests 3 and 4 instead of the actual tests, for two reasons
* It's easier to see the steps
* It's easier to avoid setting up policies every time without adding unecessary parameters in the existing tests

Example of running the script
```
./big-data-c3-tests/trino-spark-tests/debugging_scripts/test_spark_drop_error.sh /home/projects true 5 showDbsAfter
```


trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/debugging_scripts/test_spark_drop_error.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
```
xBis7, line 1: @GeorgeJahad Maybe rename the script to `debug_spark_drop_db_error.sh`?

GeorgeJahad: I would call it debug_no_such_namespace_error.sh

GeorgeJahad (approved):
This is good enough for now.  Longer term we may want to create more generic scripts and then build tools on top of that, but that isn't important enough yet.

thanks!

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/debugging_scripts/test_spark_drop_error.sh
xBis7: Done.

PR #55, opened 2024-07-17
Confirm WRITE and CREATE failures for automated tests

xBis7:


trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -268,9 +268,11 @@ runTrino() {
   echo ""
 
   if [ "$CURRENT_ENV" == "local" ]; then
-    docker exec -it -u "$user" "$TRINO_HOSTNAME" trino --execute="$trino_cmd" 2>&1 | tee "$TRINO_TMP_OUTPUT_FILE"
+    # Use '--output-format ALIGNED' to make sure that there is always output and it has the correct format.
+    # This option, aligns the output columns in tabs.
+    docker exec -it -u "$user" "$TRINO_HOSTNAME" trino --output-format ALIGNED --execute="$trino_cmd" 2>&1 | tee "$TRINO_TMP_OUTPUT_FILE"
```
xBis7, line 273: `--output-format ALIGNED` This addition is because in a query that doesn't return a result and the output is `(0 rows)`, it doesn't print anything.

If we run the command directly in trino-shell then it's ok, but there is no output when using `docker exec`. This fixes it but I don't have a way of confirming whether this works in c3 as well.

xBis7:
I'm looking into better error handling in case of failure. This is some example output where the test failed.

```
=> Testing that 'createTable' has failed as expected.

24/07/18 11:05:47 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Spark context available as 'sc' (master = spark://spark-master:7077, app id = app-20240718110547-0002).


Running command:
spark.sql("select * from gross_test.test2").show

Expecting it to fail.
--------------------------


Hive Session ID = 90916fe8-3e04-4d47-957e-a6ad979a492e
+---+----+                                                                      
| id|name|
+---+----+
|  1|John|
|  2|Jane|
|  3| Bob|
+---+----+



--------------------------
Test finished without issues while it was expected to fail.
--------------------------



=> There was an error. 'createTable' was expected to have failed but it was successful.
```

I'll run all the tests again and then push the changes.

xBis7:
All tests passed. I've updated the PR.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -293,6 +295,131 @@ runTrino() {
   fi
 }
 
+# -- CHECK CREATE/WRITE FAILURES --
+verifyCreateWriteFailure() {
+  # Run all checks as user1 for now.
```
GeorgeJahad, line 300: Is this because user1 has more permissions? If so, then please add that to the comment.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -293,6 +295,131 @@ runTrino() {
...
+  component=$1
+  operation=$2
```
GeorgeJahad, line 302: Please list the possible values here.  We haven't been doing a very good job of documenting parameters.  Let's start doing a better job of it.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -293,6 +295,131 @@ runTrino() {
...
+  db_name=$3
+  table_name=$4
+  insert_id=$5 # All tables have an id field.
```
GeorgeJahad, line 305: I know what you mean here, but to me it is clearer to say:
"All rows have an id value passed in during the insert."

also if this is an optional parameter please make that clear.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -293,6 +295,131 @@ runTrino() {
...
+    if [ "$component" == "spark" ]; then
+      command="spark.sql(\"show tables in $db_name\").show"
+      expectedOutput="Database '$db_name' not found"
+
+      runSpark "$SPARK_USER1" "$command" "shouldFail" "$expectedOutput"
+    else
+      command="show tables in $TRINO_HIVE_SCHEMA.$db_name"
+      expectedOutput="Schema '$db_name' does not exist"
+
+      runTrino "$TRINO_USER1" "$command" "shouldFail" "$expectedOutput" "user"
```
GeorgeJahad, line 326: USER1 shouldn't have the final "user" parameter

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
```diff
@@ -293,6 +295,131 @@ runTrino() {
...
+    if [ "$component" == "spark" ]; then
+      command="spark.sql(\"show tables from $db_name\").show"
+      expectedOutput="$table_name"
+
+      runSpark "$SPARK_USER1" "$command" "shouldPass" "$expectedOutput"
+    else
+      command="show tables in $TRINO_HIVE_SCHEMA.$db_name"
+      expectedOutput="$table_name"
+
+      runTrino "$TRINO_USER1" "$command" "shouldPass" "$expectedOutput" "user"
```
GeorgeJahad, line 368: same as above

GeorgeJahad:
why are renames not checked?

spark tests 10 and 11 and trino tests 9 and 11 all have alters.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_9.sh
```diff
@@ -45,7 +45,10 @@ expectedMsg="Permission denied: user [$TRINO_USER1] does not have [SELECT] privi
 # 2nd parameter: the command to be executed
 # 3rd parameter: 'shouldPass' if the command should succeed and 'shouldFail' if the command should fail
 # 4th parameter: the expected output message. For Trino all commands (whether successful or not) have an expected output message.
-runTrino "$TRINO_USER1" "$command" "shouldPass" "$expectedMsg"
+runTrino "$TRINO_USER1" "$command" "shouldFail" "$expectedMsg"
+
+# We can't check if the create failed here because there is no READ access for any user.
+# verifyCreateWriteFailure "trino" "createDb" "gross_test"
```
GeorgeJahad, line 51: > # We can't check if the create failed here because there is no READ access for any user.


Can we just change permissions temporarily?

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
xBis7: That `"user"` parameter has something to do with c3. I'm not sure what is used for but because @mladjan-gadzic added it in the other calls, I added it here as well.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
xBis7: Yes, that's correct. I'll update the comment.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
xBis7: I've updated the comment.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_9.sh
xBis7: @GeorgeJahad It turns out that although the schema creation should have failed and it gave us an exception, the create worked. `verifyCreateWriteFailure` showed that the schema was actually created.

```
=> Testing that 'createDb' has failed as expected.


==========================
Running trino command: 'show tables in hive.gross_test' for user 'trino' 
--------------------------

 Table 
-------
(0 rows)

--------------------------
The command failed. It didn't have the expected output.
Expected output: 'Schema 'gross_test' does not exist'
Actual output: ' Table 
-------
'0 rows)
==========================
```

I'll look more into this.

xBis7:
> why are renames not checked?
> 
> spark tests 10 and 11 and trino tests 9 and 11 all have alters.

@GeorgeJahad I've extended the code to check for rename failures. I've run all the tests and there are no issues.

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_9.sh
xBis7: I just realized that `test_7.sh` creates the schema. Here we are expecting to get a SELECT error while trying to create a schema that already exists without having enough permissions to read and figure out that it's there.

IMO, we should add a comment here saying that and skip the check.

xBis7: I added a comment explaining that the db already exists from `test_7.sh`.

xBis7:
@GeorgeJahad I've updated the PR. I'm not sure about the `"user"` parameter in `runTrino`.

trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh
xBis7: I cleaned it up.

GeorgeJahad (approved):
lgtm

xBis7:
* rebased the branch
* fixed the conflicts with master
* reran the tests, all are still passing

I'm merging this.

@GeorgeJahad Thanks for the help!

PR #57, opened 2024-07-23
Set default HDFS dir permissions to 700

xBis7:
In this PR, we are setting this HDFS config property `fs.permissions.umask-mode=077` which makes sure that the default posix permissions for new paths, will be
* 700 for directories
* 600 for files

Further changes
* `changeHdfsPathPermissions`
  * has been renamed from `changeHdfsDirPermissions`
  * the order of the parameters has been changed so that it's easier to read
* Permissions for `/opt/hive/data` have been set to 755
* Permissions for `/data/projects` have been set to 755
* Permissions for HDFS test file `test.csv` have been set to 655

All tests, 
* automated
* load
* spark internal
* trino internal

are passing.


GeorgeJahad:
> All tests,
>
>    automated

All of our tests are "automated" so let's start calling these the "c3" tests

trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/spark/test_3.sh
```diff
@@ -30,3 +30,7 @@ command="spark.sql(\"create database gross_test location '/data/projects/gross_t
 # 3rd parameter: 'shouldPass' if the command should succeed and 'shouldFail' if the command should fail
 # 4th parameter: the expected error message if the previous parameter is 'shouldFail'
 runSpark "$SPARK_USER1" "$command" "shouldPass"
+
+# Update permissions to the parent dirs.
```
GeorgeJahad, line 34: shouldn't these be in the setup.sh script?  also these should just be set locally not in c3, right?

xBis7: We don't explicitly create dirs `/data/projects` but spark does it for us when we run `create database`.

I don't know about c3 whether it's the same case or they create them in advance and set the permissions.

xBis7: I'll move these changes out of this script.

trino-hms-hdfs-ranger/tests/test_hdfs_data_creation.sh
```diff
@@ -30,3 +30,9 @@ retryOperationIfNeeded "$abs_path" "createHdfsFile $HDFS_DIR" "$notExpMsg" "fals
 
 notExpMsg="Permission denied"
 retryOperationIfNeeded "$abs_path" "createHdfsDir $HIVE_WAREHOUSE_DIR" "$notExpMsg" "false" "true"
+
```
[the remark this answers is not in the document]
xBis7: Good idea. Done.

trino-hms-hdfs-ranger/testlib.sh
```diff
@@ -1036,6 +1036,21 @@ base64encode() {
   fi
 }
 
+setupHdfsPathsAndPermissions() {
+  createHdfsDir "$HDFS_DIR"
+  createHdfsFile "$HDFS_DIR"
+  changeHdfsPathPermissions "$HDFS_DIR" 755
+  changeHdfsPathPermissions "$HDFS_DIR/test.csv" 655
+
```
GeorgeJahad, line 1044: I can' t remember any more why these are not 700/600.  If there is a reason why they should not be, please add a comment

GeorgeJahad (approved):
one last nit, and then we should be good to merge

trino-hms-hdfs-ranger/testlib.sh
xBis7: This setup is specific to our internal tests. It isn't used in the c3 tests. In our internal tests, we don't want to get an EXECUTE error. I added a comment.

xBis7: This setup is specific to our internal tests. It isn't used in the c3 tests. In our internal tests, we don't want to get an EXECUTE error. I added a comment.

xBis7:
@GeorgeJahad Thanks for the review!

PR #59, opened 2024-07-25
Reuse the c3 lib methods throughout all tests

xBis7:
There were a lot of methods that were duplicated and copied for the `testlib.sh` to the c3 `lib.sh` and the other way around. 

In addition, in case we made a change to the c3 tests and wanted to apply that change to the rest of the tests, then we would have to rewrite the method under `testlib.sh`.

To avoid maintaining the same code logic in multiple places, I've simplified `testlib.sh` and removed all the methods that also exist in `c3/lib.sh`. I think the approach in `c3/lib.sh` for running the tests, is cleaner and simpler and that's why I chose those methods over the ones in `testlib.sh`.

I've refactored all the tests to use the same methods that the c3 tests are using. Now every change to the c3 tests will be automatically applied to the rest of the tests.

I've run the tests multiple times and they are passing without issues.

There are no functional changes in this PR.


GeorgeJahad:
This is a huge pr.  please break it down into 2 or 3 pieces and add more explanations as to what is being done.

xBis7:
This PR has been split into new smaller PRs that have been merged. Closing.

PR #61, opened 2024-07-26
Extend internal tests to also check for create/write failures

xBis7:
This PR is a draft because it's depending on PR https://github.com/xBis7/docker-setup-helper-scripts/pull/59.

I've run all tests multiple times and they are all passing.


trino-hms-hdfs-ranger/tests/spark/4_test_spark_only_select_perm.sh
```diff
@@ -38,14 +40,25 @@ command="spark.sql(\"alter table $TABLE_ANIMALS drop partition (name='cow')\")"
 expectedMsg="Permission denied: user [spark] does not have [ALTER] privilege on [$DEFAULT_DB/$TABLE_ANIMALS]"
 runSpark "spark" "$command" "shouldFail" "$expectedMsg"
 
+# We haven't added any data to the partition and therefore it won't show up in 'describe table'.
```
GeorgeJahad, line 43: is there a reason not to add data to the partition?

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/4_test_alter_table_uri.sh
```diff
@@ -37,6 +37,34 @@ fi
...
+#
+# spark.sql("describe table gross_test.gross_test_table").show(true)
+# +--------+---------+-------+
+# |col_name|data_type|comment|
+# +--------+---------+-------+
+# |      id|      int|   null|
+# |     num|      int|   null|
+# +--------+---------+-------+
+#
+# We can't check the location
```
GeorgeJahad, line 64: can't you just do an ls on the new location?

trino-hms-hdfs-ranger/tests/trino/1_test_trino_no_hdfs_perms.sh
```diff
@@ -6,7 +6,7 @@ set -e
 
 updateHdfsPathPolicy "/*" "read,write,execute:hadoop"
 updateHiveDbAllPolicy "*" "select,update,Create,Drop,Alter,Index,Lock,All,Read,Write,ReplAdmin,Refresh:hive"
-updateHiveDefaultDbPolicy "select,read:spark"
+updateHiveDefaultDbPolicy "select,read:spark,trino"
```
GeorgeJahad, line 9: why did this change?

trino-hms-hdfs-ranger/tests/trino/1_test_trino_no_hdfs_perms.sh
```diff
@@ -16,7 +16,9 @@ echo "- INFO: Trino user trino shouldn't be able to create a table without HDFS
 echo "- INFO: All policies are to their defaults and Hive access to default DB has been removed for group public."
 echo ""
 
-command="create table hive.$DEFAULT_DB.$TRINO_TABLE (column1 varchar,column2 varchar) with (external_location = 'hdfs://namenode/$HDFS_DIR',format = 'CSV');"
+command="create table hive.$DEFAULT_DB.$TRINO_TABLE (id varchar, name varchar) with (external_location = 'hdfs://namenode/$HDFS_DIR',format = 'CSV');"
```
GeorgeJahad, line 19: why did this change?

trino-hms-hdfs-ranger/tests/trino/2_test_trino_hdfs_perms_no_hive_perms.sh
```diff
@@ -21,7 +21,15 @@ waitForPoliciesUpdate
...
 expectedMsg="Permission denied: user [trino] does not have [CREATE] privilege on"
 runTrino "trino" "$command" "shouldFail" "$expectedMsg"
+
+# Provide select access to verify that the table creation failed.
+updateHiveDefaultDbPolicy "select:trino"
+waitForPoliciesUpdate
+
+verifyCreateWriteFailure "trino" "createTable" "$DEFAULT_DB" "$TRINO_TABLE"
+
+# The next test will set new policies. No need to reset them here as well.
```
GeorgeJahad, line 35: please reset them back.  It is better to be consistent, and not have to think about what the next test does.

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
```diff
@@ -20,7 +20,7 @@ echo "- INFO: Select from $TRINO_TABLE table."
 echo "- INFO: [select] should succeed."
 
 command="select * from hive.$DEFAULT_DB.$TRINO_TABLE"
-expectedMsg="1       |  dog"
+expectedMsg="dog"
```
GeorgeJahad, line 23: why did this change?

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
```diff
@@ -38,9 +41,13 @@ command="alter table hive.$DEFAULT_DB.$TRINO_TABLE rename to $NEW_TRINO_TABLE_NA
...
+verifyCreateWriteFailure "trino" "renameTable" "$DEFAULT_DB" "$TRINO_TABLE"
+
 echo ""
 echo "- INFO: Insert into $TABLE_ANIMALS table."
 echo "- INFO: [alter] should fail."
 command="insert into hive.default.$TABLE_ANIMALS values (1, 'cat');"
 expectedMsg="Permission denied: user=trino, access=WRITE, inode=\"/\":hadoop:supergroup:drwxr-xr-x"
 runTrino "trino" "$command" "shouldFail" "$expectedMsg"
+
+verifyCreateWriteFailure "trino" "insertInto" "$DEFAULT_DB" "$TABLE_ANIMALS" "1"
```
GeorgeJahad, line 53: This should be consistent with the command on line 49.  Either change this to "default" or change that to "$DEFAULT_DB"

trino-hms-hdfs-ranger/tests/spark/4_test_spark_only_select_perm.sh
xBis7: Something is wrong with the way the table has been created. I can add data to the table but when I run 

```
scala> spark.sql("describe extended table animals").show(false)
org.apache.spark.sql.AnalysisException: Table or view not found: table; line 1 pos 18;
'DescribeColumn 'animals, true, [info_name#82, info_value#83]
+- 'UnresolvedTableOrView [table], DESCRIBE TABLE, true
```

but the table is there
```
scala> spark.sql("describe table animals").show
+--------------------+---------+-------+
|            col_name|data_type|comment|
+--------------------+---------+-------+
|                  id|      int|   null|
|                name|   string|   null|
|# Partition Infor...|         |       |
|          # col_name|data_type|comment|
|                name|   string|   null|
+--------------------+---------+-------+
```

I'm looking into this. I think that some data like the location is missing from the create and that's why it's failing.

xBis7: The command is wrong
```
spark.sql("describe extended table animals").show(false)
```

The correct command is 
```
spark.sql("describe extended animals").show(false)
```

I found only two usages and in both of them the call has been commented out. That's why I didn't catch this earlier.

xBis7: Fixed!

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
xBis7: You are right. I changed it and cleaned up all other occurences.

trino-hms-hdfs-ranger/tests/trino/1_test_trino_no_hdfs_perms.sh
xBis7: I added `select` access so that the trino user can do a read to verify that `createTable` failed. It doesn't make a difference for this specific test since we are testing for HDFS access.

trino-hms-hdfs-ranger/tests/trino/1_test_trino_no_hdfs_perms.sh
xBis7: I changed it for consistency. All other tables have an id column and I'm using the id to verify whether the data is there or not.

Also, if you see in test 5 where we insert data, it's inserting an id string.

```
command="insert into hive.$DEFAULT_DB.$TRINO_TABLE values ('5', 'cat');"
```

trino-hms-hdfs-ranger/tests/trino/2_test_trino_hdfs_perms_no_hive_perms.sh
xBis7: Done.

trino-hms-hdfs-ranger/tests/trino/4_test_trino_only_select_perm.sh
xBis7: https://github.com/xBis7/docker-setup-helper-scripts/blob/master/trino-hms-hdfs-ranger/big-data-c3-tests/lib.sh#L271-L273

We have added the option `--output-format ALIGNED` to make the output more reliable. I think that in some cases, part of the output was malformed. But this is causing the number of spaces between columns to change depending on the screen.

Without the option, the output here is printed as `"1" "," "dog"` and with the option, it's printed in columns but the number of spaces isn't constant.

trino-hms-hdfs-ranger/tests/spark/hive_url_policies/4_test_alter_table_uri.sh
xBis7: This turned out to be an issue with my code.

This command is wrong
```
spark.sql("describe extended table gross_test.gross_test_table").show
```

after fixing it I just uncommented the code and it works.

xBis7:
@GeorgeJahad I've addressed all your comments.

GeorgeJahad (approved):
lgtm, thanks!

PR #62, opened 2024-07-26
Extend trino/test_9 to test a deny write policy

xBis7:


trino-hms-hdfs-ranger/big-data-c3-tests/trino-spark-tests/trino/test_9.sh
```diff
@@ -52,6 +52,28 @@ runTrino "$TRINO_USER1" "$command" "shouldFail" "$expectedMsg"
 # We are testing that we don't have enough permissions to perform a read and see that
 # the db exists.
 
+# Provide select access.
```
GeorgeJahad, line 55: # Provide select access so we can see the deny error.

GeorgeJahad (approved):
lgtm

PR #63, opened 2024-07-29
Reuse-c3_lib.sh-methods-1: Replace existing policy methods

xBis7:
This is the 1st sub-PR from

https://github.com/xBis7/docker-setup-helper-scripts/pull/59

It's replacing all methods for setting Ranger policies. The main change is the order of the method parameters. The resources come first.

I've also removed the `base64encode` to avoid a conflict. It has the same method body, there are no further changes needed.

### DIFF
--- a/trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
+++ b/trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 source "./testlib.sh"
+source "./big-data-c3-tests/lib.sh"
 
 set -e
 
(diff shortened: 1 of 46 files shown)


trino-hms-hdfs-ranger/setup/setup_for_testing_hms_policies.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 source "./testlib.sh"
+source "./big-data-c3-tests/lib.sh"
```
GeorgeJahad, line 4: is there a reason why testlib.sh doesn't source this? instead of adding it to every file?

xBis7: You are right. I'll update the patch.

GeorgeJahad (approved):
lgtm.  thanks!

PR #64, opened 2024-07-29
Reuse-c3_lib.sh-methods-2: Replace existing hdfs methods

xBis7:
This is based on the branch of PR: https://github.com/xBis7/docker-setup-helper-scripts/pull/63

It's replacing all HDFS methods that already exist in c3/lib.sh.

PR #65, opened 2024-07-29
Reuse-c3_lib.sh-methods-3: Replace spark methods

xBis7:
This PR is based on the branch of PR: https://github.com/xBis7/docker-setup-helper-scripts/pull/64

This patch has the following changes
* The scala files used for running spark tests have been deleted
* All calls have been replaced with `runSpark`
* The scala file from the c3 tests is used
* During setup we need to copy the new scala file under the spark container


trino-hms-hdfs-ranger/tests/spark/test_for_exception.scala
```diff
@@ -1,15 +0,0 @@
-val sqlStr = spark.conf.get("spark.app.sql", "default_sql_value")
```
GeorgeJahad, line 1: @xBis7 @mladjan-gadzic 

I thought these files were created because we couldn't check the exit code on the spark shell.  Am I not remembering that correctly.

GeorgeJahad: ah, I guess they have been replaced by:
big-data-c3-tests/trino-spark-tests/spark/scala-test-files/TestCmd.scala

GeorgeJahad (approved):
lgtm

trino-hms-hdfs-ranger/tests/spark/test_for_exception.scala
xBis7: Yes, that's correct. We are still using scala to check the exit code.

